PR #10, opened 2017-11-16
Enable simple travis build, add missing code-coverage m4

grondo:
Add the missing m4 macro for code coverage and enable travis-ci sanity check of the flux-security build so that this NEVER HAPPENS AGAIN.


garlick:
Yay!

Issue #5, opened 2017-11-13
Define flux-security configuration file format


grondo:
The flux-security project will require one or more configuration files to allow site control of features. At the very least, the IMP may read system configuration to determine what users and/or groups are allowed to run a privileged version of the IMP in multi-user mode.

The configuration file format should be kept very simple to reduce the amount of code and potential for vulnerability during configuration parsing.

grondo:
For flux-security purposes we probably only need key/value configuration file at first. Something like INI would work fine for this, and there are some very small ini libraries would could use out there:

  https://github.com/compuphase/minIni
  https://github.com/benhoyt/inih

Since we have to parse YAML anyway in the IMP at least, we could just use YAML for configuration files (like everyone else). This would allow more powerful configuration in the future, and is a nice, human-editable format.

Another interesting spec along these same lines is [TOML](https://github.com/toml-lang/toml), which is meant to be human readable/editable, but much simpler than YAML.

garlick:
I've been using [this INI parser](https://github.com/benhoyt/inih) in a few of my personal projects.  Very simple.

grondo:
Thanks, that seems to be the same as the second project referenced above.

If we want to standardize on config file format I'm not sure INI is sufficient. (doesn't support complex values). YAML might be more flexible on that front..

garlick:
Maybe the same config file format could be used for certificates as outlined  (#7)

TOML looks pretty neat.

garlick:
One interesting thing about TOML is that it converts neatly to JSON.  Presumably you could call something like `toml_to_jansson(FILE)` and just use the familiar jansson API to access your config stuff...

Come to think of it, a JSON library also gives you hashes and arrays, which might be an alternative to pulling in a list + hash implementation, like inevitably happens...

garlick:
I think we've got all this mostly nailed down, so closing.

PR #11, opened 2017-11-17
libtoml99:  add TOML parser library

garlick:
As discussed in #5, we need a config file and [TOML](https://github.com/toml-lang/toml) seems like a good one to try.  I wanted to try changing the "zcert" format certificates we used in the IMP prototype to TOML also, to get J signing bootstrapped.

The above page lists two C implementations as supporting the latest TOML spec (0.4.0):

[libtoml](https://github.com/ajwans/libtoml) requires  ragel (the parser generator) and libicu for unicode support.  It doesn't provide accessors for TOML types - all values are accessed as strings.  Finally, it has a lot of unchecked asprintf calls and likes to print to stderr from within the parser function.

[tomlc99](https://github.com/cktan/tomlc99) has no dependencies, provides accessors for TOML types, and seems cleaner.  One downside is that it seems to be a personal project with 6 vaguely commented commits in the repo.

Anyway this PR proposes to add the latter.


grondo:
Thanks, this seems fine and we can always update to a better libtoml later if the TOML config files even work out.

Also, sorry about forgetting to `git add DISCLAIMER.LLNS`

Also, nice job adding tests.

Issue #7, opened 2017-11-15
need certificate format


garlick:
It's useful to have a certificate format to store user's public and private keys.

In our prototype imp, we used a derivative of czmq's [zcert](http://czmq.zeromq.org/manual:zcert) certificates, but those are dependent on libczmq and we probably don't want to link with that.  The certs look like this

public key:
```
#   ****  Generated on 2017-04-05 13:37:15 by CZMQ  ****
#   ZeroMQ CURVE Public Certificate
#   Exchange securely, or use a secure mechanism to verify the contents
#   of this file after exchange. Store public certificates in your home
#   directory, in the .curve subdirectory.

metadata
    time = "2017-04-05T13:37:15"
    role = "signature"
curve
    public-key = "fyAQm1uvTk!SPm@cN2=U5W1mJJmK3{X:@N?g(t4W"
```

private key (secret-key redacted):
```
#   ****  Generated on 2017-04-05 13:37:15 by CZMQ  ****
#   ZeroMQ CURVE **Secret** Certificate
#   DO NOT PROVIDE THIS FILE TO OTHER USERS nor change its permissions.

metadata
    time = "2017-04-05T13:37:15"
    role = "signature"
curve
    public-key = "fyAQm1uvTk!SPm@cN2=U5W1mJJmK3{X:@N?g(t4W"
    secret-key = "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
```

garlick:
TOML seems like a nice fit here, e.g. the above would be
```toml
[metadata]
    time = 2017-04-05T13:37:15
    role = "signature"
[curve]
    public-key = "fyAQm1uvTk!SPm@cN2=U5W1mJJmK3{X:@N?g(t4W"
```
```toml
[metadata]
    time = 2017-04-05T13:37:15
    role = "signature"
[curve]
    public-key = "fyAQm1uvTk!SPm@cN2=U5W1mJJmK3{X:@N?g(t4W"
    secret-key = "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"

garlick:
One idea we had discussed was using an IMP key-pair to sign/verify user public key certs, so they can be sent along with signed requests.  The advantage is then verification only needs access to the IMP public key and a revocation list, not the full set of user public keys.

[...]

PR #13, opened 2017-11-17
libtomlc99/test: fix remaining test problems

garlick:
This PR adds a fix to libtomlc99 backported from upstream,
and improves our TAP test so that we no longer need to blacklist
any of the BurntSushi test input.

### DIFF
--- a/src/libtomlc99/toml.c
+++ b/src/libtomlc99/toml.c
@@ -1140,6 +1140,10 @@ static void parse_select(context_t* ctx)
 	/* [[x.y.z]] -> create z = [] in x.y */
 	toml_array_t* arr = create_keyarray_in_table(ctx, ctx->curtab, z,
 						     1 /*skip_if_exist*/);
+        if (!arr) {
+            e_syntax_error(ctx, z.lineno, "key exists");
+            return;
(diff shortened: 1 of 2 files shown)


garlick:
Don't merge yet @grondo ... hang on.

garlick:
OK rebased after amending commits on master.

garlick:
The author of tomlc99 fixed the segfault I reported, so adding that fix too.
All BurntSushi tests are now un-blacklisted.

This is OK to merge as far as I'm concerned.

dongahn:
This is closed but I have a question. I believe sched also need a configuration file that allows us to specify system parameters. (e.g., the default duration if a jobspec does not have one; default exclusivity of a type if the spec unspecified the exclusive field on that type ect.) I wasn't following how the core code that is using TOML. Could one of you please point me to the code?

grondo:
No other flux-framework projects are using TOML at this point. I think using it here first might be a low risk method to determine suitability/usability of TOML for a standardized config file format for all framework projects.

Example code should show up in this repo within the next couple weeks if you want to take a look. Also see issues #5 and #7.

PR #15, opened 2017-11-21
add certificate class for libsodium based sign/verify

garlick:
This is similar to the "zsigcert" class we used for JSON signing in the IMP prototype, except it is not dependent on czmq.  It uses TOML for the certificate format.

This is very bare bones.  Basically you can
* create a cert with new keys
* save the cert to TOML files (name and name.pub)
* load the cert from a TOML file
* compare two certs (compare their keys, that is)
* create a detached (base64) signature for an arbitrary blob
* verify a detached (base64) signature for an arbitrary blob

Major missing functionality (to be added later)
* JSON sign/verify
* CA certificate sign/verify
* support for data other than keys in the cert

I was hoping maybe we could iterate on basic stuff (like is this the right abstraction, name changing, etc) in this PR and then do the other stuff in another PR.  I don't have a lot invested in this particular design so happy to discuss completely different alternatives if you like.
```c
struct flux_sec_sigcert;

/* Destroy cert.
 */
void flux_sec_sigcert_destroy (struct flux_sec_sigcert *cert);

/* Create cert with new keys
 */
struct flux_sec_sigcert *flux_sec_sigcert_create (void);

/* Load cert from name, falling back to name.pub
 */
struct flux_sec_sigcert *flux_sec_sigcert_load (const char *name);

/* Store cert to name and name.pub
 */
int flux_sec_sigcert_store (struct flux_sec_sigcert *cert, const char *name);

/* Return true if two certificates have the same keys.
 */
bool flux_sec_sigcert_equal (struct flux_sec_sigcert *cert1,
                             struct flux_sec_sigcert *cert2);

/* Return a base64 signature over buf, len.
 * Caller must free.
 */
char *flux_sec_sigcert_sign (struct flux_sec_sigcert *cert,
                             uint8_t *buf, int len);

/* Verify a base64 signature over buf, len using cert.
 * Returns 0 on success, -1 on failure.
 */
int flux_sec_sigcert_verify (struct flux_sec_sigcert *cert,
                             const char *signature, uint8_t *buf, int len);
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,12 @@ addons:
   apt:
   packages:
 
+before_install:
+  - curl -L -O --insecure https://github.com/jedisct1/libsodium/releases/download/1.0.10/libsodium-1.0.10.tar.gz
+  - tar -xf libsodium-1.0.10.tar.gz
+  - cd libsodium-1.0.10 && ./configure --prefix=$HOME/local && make && make install
+  - export PKG_CONFIG_PATH=$HOME/local/lib/pkgconfig:$HOME/local/share/pkgconfig
+
 script:
   - ./autogen.sh
   - ./configure
--- a/configure.ac
+++ b/configure.ac
@@ -62,6 +62,10 @@ LT_INIT
 #
 AC_HEADER_STDC
 
+#
+#  Checks for packages
+#
+PKG_CHECK_MODULES([SODIUM], [libsodium], [], [])
 
 #
 #  Other checks
--- a/src/lib/Makefile.am
+++ b/src/lib/Makefile.am
@@ -5,4 +5,12 @@ AM_CFLAGS = \
 AM_LDFLAGS = \
 	$(CODE_COVERAGE_LIBS)
 
-AM_CPPFLAGS =
+AM_CPPFLAGS = \
+	-I$(top_srcdir)
+
+noinst_LTLIBRARIES = \
+	libflux-security.la
+
+libflux_security_la_SOURCES = \
+	sigcert.c \
+	sigcert.h
--- a/src/lib/Makefile.am
+++ b/src/lib/Makefile.am
@@ -1,12 +1,15 @@
 AM_CFLAGS = \
 	$(WARNING_CFLAGS) \
+	-Wno-unused-parameter \
 	$(CODE_COVERAGE_CFLAGS)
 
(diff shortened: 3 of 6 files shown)


garlick:
Argh, libsodium-dev is not whitelisted for travis-ci.

https://github.com/travis-ci/apt-package-whitelist/issues/4047

grondo:
Sorry about missing libsodium! Maybe we should try to move CI to buildbot for this repo sooner rather than later!

grondo:
Nice work thanks!

I wonder if we need the `flux_sec_*` prefix on *all* flux-security API calls? Would `flux_sigcert` be sufficient?

garlick:
Yes that works for me.

What do you think about the lack of typedef (use of `struct flux_sigcert` instead of `flux_sigcert_t`)?

grondo:
> What do you think about the lack of typedef (use of struct flux_sigcert instead of flux_sigcert_t)?

I could go either way. I figured you probably had a reason for dropping the typedef.

garlick:
No reason really.  I've over-used them in the past.  Here though it might be the right thing.

grondo:
Ok, works for me.

garlick:
Just forced pushed a class rename.  In retrospect I probably shouldn't have squashed that.  Hope I didn't screw up any reviewing you were doing @grondo!

grondo:
No problem on the squash, at this point I wasn't reviewing the code too carefully at this point.

grondo:
This good to go in?

garlick:
I have a couple of tweaks: some comment typos and checking for one more kind of invalid input, with tests.  I'll go ahead and squash/push then OK.

I didn't add the typedef.  I dunno, seems OK this way.

garlick:
Ready now if you like.

grondo:
> I didn't add the typedef. I dunno, seems OK this way.

Seems fine to me too.

One minor thing which is probably just a personal preference: 

IMO, though the commit subject is probably descriptive enough, commit 32ab237 could use a message body introducing the purpose of the sigcert class and simple description (just when going through git-log history it can add a bit of context)

garlick:
Good idea!  Let me add that.

garlick:
OK, I added some explanation to that commit.

grondo:
Nice thanks! Merge?

garlick:
Sure.

PR #17, opened 2017-11-21
travis-ci: add gcc 5, 7 and clang 4 builds and include code coverage

grondo:
This PR first fixes the accidentally broken travis-ci build, then adds support for GCC 5, 7, and clang 4.0 compilers. Finally, code coverage is added via coveralls just so we begin tracking coverage early (and because it was easy)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,25 @@ compiler:
   - gcc
   - clang
 
+matrix:
+  include:
+    - compiler: gcc-5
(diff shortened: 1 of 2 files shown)


garlick:
Oops sorry for my sloppiness with the libsodium cwd.

Should we be running distcheck here?

grondo:
> Should we be running distcheck here?
Yeah, maybe in one build. That will catch some errors probably (haven't even tried it yet.)

grondo:
Thanks, just pushed try 1. Not sure if it will work or not.

garlick:
Great!  Ready to merge?

grondo:
Think so

grondo:
Thanks!

Issue #16, opened 2017-11-21
need keygen utility for signing keys


garlick:
We need something similar to `flux keygen` for creating/updating signing keys.

Did we settle on how we wanted to add commands in this project?

If we want certs to be signed by the IMP, such a command would require access to the IMP's private signing key.

grondo:
Yeah, seems like a CA type role would have to be an IMP subcommand, but you shouldn't need to run a setuid binary to create a pair of signing keys.

Should `flux-keygen` remain part of flux-core, but require code from libflux-security.so in order to create signing keys? This would allow even flux-core in a builddir to correctly use the system security library to create/manage keys. If we try to make `keygen` a `flux(1)` subcommand installed by flux-security project, things would seem to get messy real quick

garlick:
Having `flux-keygen` reside in flux-core sounds good.

So if we had a `flux-imp ca` subcommand to provide a "CA signing service", would we provide API calls (such as new methods in sigcert) to interact with it, optionally?

garlick:
Backing up, I was picturing that the sigcert class could have methods like
```c
int flux_sigcert_sign_cert (struct flux_sigcert_t *ca_cert,
                            struct flux_sigcert_t *cert);

int flux_sigcert_verify_cert (flux_sigcert_t *ca_cert,
                              struct flux_sigcert_t *cert);
```

These would sign `cert` with `ca_cert`, adding the signature to `cert` and verify `cert` with `ca_cert`, getting the signature from `cert`.  These would use internal methods to serialize `cert` for signing/verification purposes.

So the caller to `sign_cert` would need to provide the CA's private cert.  The caller to `verify_cert` would need to provide the CA's public cert, and also potentially check a list of revocations. 

It's feasible that `verify_cert` could be wrapped in another API call that parses an IMP config file and locates the CA's public cert and revocation list, which could be publicly readable.  This is probably a good way to go since it would need to be used wherever J needs to be validated if public certs are provided along with the signed J.

`sign_cert` needs access to the CA's private cert which cannot be publicly readable.  This could similarly be wrapped in an API call that parses config and locates the CA's private cert, but it could only be called from a process with read permission on the CA's private cert.

If `flux-keygen` lives in flux-core, it seems like it needs some way to call out to a service or executable in `flux-security` to do the CA signing.  Or maybe it should simply exec the IMP and outsource the whole job to this project?

Did you have any thoughts @grondo?

grondo:
[...]

PR #18, opened 2017-11-22
sigcert: add json_loads/dumps methods

garlick:
This adds `flux_sigcert_json_loads()` and `flux_sigcert_json_dumps()` to load/dump a public version of a cert from/to a JSON string.

Eventually this could be used to include a public cert within a signed request, or to send a public cert to a CA for signing (though that requires some more discussion in #16)

This adds a jansson dependency.  Instead of requiring a newer version like flux-core and then having to build it in travis, I just allow any version and install in travis with apt.  That may be fine for this project, and I didn't want to unnecessarily complicate the travis.yml or our configure script.  We can always go there later if we need to.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ addons:
       - gcc-7
       - clang-4.0
       - lcov
+      - libjansson-dev
 
 before_install:
   - curl -L -O --insecure https://github.com/jedisct1/libsodium/releases/download/1.0.10/libsodium-1.0.10.tar.gz
--- a/configure.ac
+++ b/configure.ac
@@ -66,6 +66,7 @@ AC_HEADER_STDC
 #  Checks for packages
 #
 PKG_CHECK_MODULES([SODIUM], [libsodium], [], [])
+PKG_CHECK_MODULES([JANSSON], [jansson], [], [])
 
 #
 #  Other checks
--- a/src/lib/Makefile.am
+++ b/src/lib/Makefile.am
@@ -5,11 +5,11 @@ AM_CFLAGS = \
 
 AM_LDFLAGS = \
 	$(CODE_COVERAGE_LIBS) \
-	$(SODIUM_LIBS)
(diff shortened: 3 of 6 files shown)


grondo:
Just so I'm sure I understand, these calls are for transmission of certs in messages, and the TOML format is meant for on-disk storage of certs only? (Sorry I might be missing something here)

garlick:
Right: messages, IMP input, etc..

An alternative would be cert -> TOML -> JSON string.

grondo:
> An alternative would be cert -> TOML -> JSON string.

I apologize, I still don't follow (definitely just me). Can you elaborate on the alternative? I thought cert was already in TOML?

From the IMP input perspective, a JSON encoded `J` might have a public cert for a guest user, including signature by the IMP or other CA? So the IMP would process the outer JSON to get the embedded signed cert as a json object, convert the object to a json string, then load it as a sigcert_t with `flux_sigcert_json_loads()`?

Sorry if the question is off-topic for this PR. I think the PR is fine, and I'm just trying to understand where in the flow of signed messages and inputs to the IMP these calls might take place.

garlick:
No this is good discussion - if we don't need this then we shouldn't merge it.

> From the IMP input perspective...

Yes that is what I was thinking.

On the alternative:  since we already can serialize a cert as TOML, we _could_ add methods to access that serialization as a string (as opposed to file oriented methods we have now), and then it could be included in a JSON object as a string, not broken out.  For the IMP input, the IMP would process outer JSON to get signed cert as a string, then load it as a sigcert_t.  Hmm, fewer steps.

I guess if there's never a need to access broken-out fields of a cert without going through the sigcert class, then maybe it would be better to just add those  TOML methods.

grondo:
Thanks, that was helpful. I think it is early enough that we should probably just merge these methods and move forward. It is easy enough to make them private later if they are only used internally, or remove them if it turns out there is another way. Sound OK?

garlick:
Thanks!

PR #20, opened 2017-11-28
lib/sigcert: allow user-defined metadata in the cert, and cleanup

garlick:
Add methods to get/set arbitrary data in a certificate, like the similar feature in CZMQ zcert.  This allows higher level class users to customize the certificate.

This PR also includes some cleanup, refactoring, and test improvement.


grondo:
Nice!

garlick:
Thanks!

PR #21, opened 2017-11-29
sigcert: support all TOML types in metadata accessors, cleanup

garlick:
Add functions to allow all the TOML types to be used in metadata sections.

Some cleanup and test coverage improvement also.

### DIFF
--- a/src/lib/sigcert.c
+++ b/src/lib/sigcert.c
@@ -244,6 +244,49 @@ int flux_sigcert_meta_getd (const struct flux_sigcert *cert,
     return 0;
 }
 
+int flux_sigcert_meta_setb (struct flux_sigcert *cert,
+                            const char *key, bool b)
+{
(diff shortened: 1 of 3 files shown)


grondo:
Nice!  One question, is the cert metadata going to be restricted to values only, or will arrays and nested tables potentially be required as well? I like the simplicity of this approach, however if we ever think nested tables might be required we might need a slightly different abstraction.

garlick:
The immediate use case I had in mind for these was a creation date, TTL, uuid, userid, etc., as might be needed for certificates signed by a CA.   I was imagining a flat list of values would be sufficient. 
 
Maybe we could move forward with this and revisit if it turns out not to be sufficient, or if working on wrapping TOML config files yields any insight on a better interface?  This one is admittedly unclever and repetitive.

garlick:
Thanks!

PR #22, opened 2017-11-30
sigcert: add certificate signing capability, cleanup

garlick:
This pr adds the ability to sign a certificate with another certificate, as preparation for CA functionality.

Some practical concerns were addressed related to getting only the public portion of a cert signed:
* "cert" and "cert.pub" content no longer overlaps (as borrowed from zcert)
* `flux_sigcert_load()` now has a boolean argument to specify whether secret portion is required
* `flux_sigcert_store()` can update "cert.pub" without updating "cert"

The signature is calculated over json-serialized metadata and public key, skipping an existing signature if present.

There is also some internal cleanup to allow keys and signatures to be manipulated and base64-encoded without needing to malloc.

Unit tests updated.

As a bonus and for inspiration, I added a toy keygen program to play with:
```
Usage: ./test_keygen [--sign=PATH] [--verify=PATH] [PATH]
```
certificates can be generated, signed, and verified.  Since certs are agnostic with respect to any kind of CA "schema", I added a couple of dumb things as a demo.  This only builds with make check and isn't meant to be anything but a toy.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,9 @@
+coverage:
+ round: down
+ range: "50...100"
+ ignore:
+ - ".*/test/.*"
+ - ".*/libtap/.*"
+
+comment:
+ layout: "header, diff, changes, tree"
--- a/src/lib/Makefile.am
+++ b/src/lib/Makefile.am
@@ -20,7 +20,9 @@ libflux_security_la_SOURCES = \
 
 TESTS = test_sigcert.t
 
-check_PROGRAMS = $(TESTS)
+check_PROGRAMS = \
+	test_keygen \
+	$(TESTS)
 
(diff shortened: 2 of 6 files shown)


garlick:
Oops had a double-free in there that was caught in travis.  Forced a push with a one line fix.

garlick:
Just tacked on more unit tests for missed coverage in the sigcert class spotted by codecov.

Not sure what to do about coverage for that toy keygen file.  I could drop it from this PR and just keep it to myself until we're ready to sort out a real keygen and/or CA scheme....

grondo:
We could add a codecov.yml to ignore files under `test/*` as well as `libtap/*` etc. See flux-core for the example.

grondo:
LGTM!

PR #23, opened 2017-12-05
Add hash to libutil and flexible logger for the IMP

grondo:
This is a placeholder PR for continuing work on the do-nothing-yet IMP.
The hash implementation from the MUNGE project is added to libutil, along with unit tests, and this implementation is used in a simplified logging interface to be used in the IMP.

The IMP now has one subcommand: `flux-imp version` as a placeholder for future work. The logging interface might not be quite correct for what we'll need (but flexible logging output is probably necessary), but this at least leaves a foundation for further work.

If it makes more sense, this PR could be pared down to just the addition of `hash.[ch]` to libutil and save the IMP work until it does something interesting (since there will be a lot of churn in the history otherwise)


garlick:
Great!

grondo:
Thanks!

PR #24, opened 2017-12-07
libuitl/cf: add config class

garlick:
This is a first cut at a config class.  It has no tests - I'm posting it for early review of API and concepts.  Here's the header for convenient reference:

```c// flags for cf_table_update
enum {
    CF_STRICT = 1,     // parse error on unknown keys
    CF_ANYTAB = 2,     // allow unknown keys if they are tables
};

// allowed types
// CF_ARRAY | <type> means array of <type> elements, where <type>
// is any type except CF_UNKNOWN, CF_TABLE, or CF_ARRAY.
enum cf_type {
    CF_UNKNOWN      = 0,
    CF_TABLE        = 1,
    CF_INT64        = 2,
    CF_DOUBLE       = 3,
    CF_BOOL         = 4,
    CF_TIMESTAMP    = 5,
    CF_STRING       = 6,
    CF_ARRAY        = 0x10,
};

struct cf_table;

struct cf_option {
    const char *key;
    enum cf_type type;
    bool required;
};
#define CF_OPTIONS_TABLE_END { NULL, 0, false }

/* Create an empty table.
 * Returns table on success, or NULL on failure with errno set.
 */
struct cf_table *cf_table_create (void);

/* Destroy a table.
 */
void cf_table_destroy (struct cf_table *cf);

/* Copy a table
 */
struct cf_table *cf_table_copy (const struct cf_table *cf);

/* cf_table_strerror() returns a descriptive error message.  It is valid when
 * cf_table_update() or cf_table_parse() return -1 with errno set to EINVAL.
 */
const char *cf_table_strerror (const struct cf_table *cf);

/* Update table 'cf' with info parsed from TOML 'buf'.
 * On success return 0.  On failure, return -1 with errno set.
 * (If errno == EINVAL, call cf_table_strerror() for descriptive error message).
 */
int cf_table_update (struct cf_table *cf, const char *buf);

/* Apply 'opts' to table 'cf' according to flags.
 * On success return 0.  On failure, return -1 with errno set.
 * (If errno == EINVAL, call cf_table_strerror() for descriptive error message).
 */
int cf_table_parse (struct cf_table *cf,
                    const struct cf_option opts[], int flags);

/* Get sub-table of 'cf' identified by 'key'.  The new table is assigned
 * to 'val' and is a standalone copy that must be freed with cf_table_destroy.
 * Returns 0 on success, or -1 on failure with errno set.
 */
int cf_table_get_table (const struct cf_table *cf,
                        const char *key, struct cf_table **val);

/* Simple accessors for values
 */
int cf_table_get_int64 (const struct cf_table *cf,
                        const char *key, int64_t *val);
int cf_table_get_double (const struct cf_table *cf,
                         const char *key, double *val);
int cf_table_get_bool (const struct cf_table *cf,
                       const char *key, bool *val);
int cf_table_get_timestamp (const struct cf_table *cf,
                            const char *key, time_t *val);
int cf_table_get_string (const struct cf_table *cf,
                         const char *key, const char **val);

/* Simple accessors for array elements.
 */
int cf_table_get_array_size (const struct cf_table *cf,
                             const char *key, int *size);
int cf_table_get_array_int64 (const struct cf_table *cf,
                              const char *key, int index, int64_t *val);
int cf_table_get_array_double (const struct cf_table *cf,
                               const char *key, int index, double *val);
int cf_table_get_array_bool (const struct cf_table *cf,
                             const char *key, int index, bool *val);
int cf_table_get_array_timestamp (const struct cf_table *cf,
                                  const char *key, int index, time_t *val);
int cf_table_get_array_string (const struct cf_table *cf,
                               const char *key, int index, const char **val);
```

### DIFF
--- a/src/lib/sigcert.c
+++ b/src/lib/sigcert.c
@@ -41,6 +41,7 @@
 
 #include "src/libtomlc99/toml.h"
 #include "src/libutil/base64.h"
+#include "src/libutil/tomltk.h"
 #include "sigcert.h"
 
 /* Define some handy types for fixed length keys and their base64 encodings.
@@ -306,30 +307,6 @@ int flux_sigcert_meta_getb (const struct flux_sigcert *cert,
     return 0;
 }
 
-/* Convert time_t (GMT) to ISO 8601 timestamp, e.g. 2003-08-24T05:14:50Z
- */
-static int timetostr (time_t t, char *buf, int size)
-{
-    struct tm tm;
-    if (!gmtime_r (&t, &tm))
-        return -1;
-    if (strftime (buf, size, "%FT%TZ", &tm) == 0)
-        return -1;
-    return 0;
-}
-
-static int strtotime (const char *s, time_t *tp)
-{
-    struct tm tm;
-    time_t t;
-    if (!strptime (s, "%FT%TZ", &tm))
-        return -1;
-    if ((t = timegm (&tm)) < 0)
-        return -1;
-    *tp = t;
-    return 0;
-}
-
 /* Timestamp is the only TOML type that doesn't have a corresponding
  * JSON type.  Represent it as a JSON object that looks like this:
  *   { "iso-8601-ts" : "2003-08-24T05:14:50Z" }
@@ -341,33 +318,24 @@ int flux_sigcert_meta_setts (struct flux_sigcert *cert,
                              const char *key, time_t t)
 {
     json_t *val;
-    char timebuf[80];
 
     if (!cert || !key || strchr (key, '.')) {
         errno = EINVAL;
         return -1;
     }
-    if (timetostr (t, timebuf, sizeof (timebuf)) < 0) {
-        errno = EINVAL;
+    if (!(val = tomltk_epoch_to_json (t)))
         return -1;
+    if (json_object_set_new (cert->meta, key, val) < 0) {
+        json_decref (val);
+        errno = ENOMEM;
     }
-    if (!(val = json_pack ("{s:s}", "iso-8601-ts", timebuf)))
-        goto nomem;
-    if (json_object_set_new (cert->meta, key, val) < 0)
-        goto nomem;
     return 0;
-nomem:
-    json_decref (val);
-    errno = ENOMEM;
-    return -1;
 }
 
 int flux_sigcert_meta_getts (const struct flux_sigcert *cert,
                              const char *key, time_t *tp)
 {
     json_t *val;
-    const char *s;
-    time_t t;
 
     if (!cert || !key || strchr (key, '.') || !tp) {
         errno = EINVAL;
@@ -377,13 +345,7 @@ int flux_sigcert_meta_getts (const struct flux_sigcert *cert,
         errno = ENOENT;
         return -1;
     }
-    if (json_unpack (val, "{s:s}", "iso-8601-ts", &s) < 0
-                                    || strtotime (s, &t) < 0) {
-        errno = EINVAL;
-        return -1;
-    }
-    *tp = t;
-    return 0;
+    return tomltk_json_to_epoch (val, tp);
 }
 
 /* encode public key 'src' to base64 'dst'
@@ -699,20 +661,8 @@ static int parse_toml_meta_set (const char *raw, struct flux_sigcert *cert,
             goto done;
     }
     else if (toml_rtots (raw, &ts) == 0) {
-        struct tm tm;
         time_t t;
-        if (!ts.year || !ts.month || !ts.day)
-            goto done;
-        if (!ts.hour || !ts.minute || !ts.second)
-            goto done;
-        memset (&tm, 0, sizeof (tm));
-        tm.tm_year = *ts.year - 1900;
-        tm.tm_mon = *ts.month - 1;
-        tm.tm_mday = *ts.day;
-        tm.tm_hour = *ts.hour;
-        tm.tm_min = *ts.minute;
-        tm.tm_sec = *ts.second;
-        if ((t = timegm (&tm)) < 0)
+        if (tomltk_ts_to_epoch (&ts, &t) < 0)
             goto done;
         if (flux_sigcert_meta_setts (cert, key, t) < 0)
             goto done;
--- a/src/libutil/Makefile.am
+++ b/src/libutil/Makefile.am
@@ -17,7 +17,9 @@ libutil_la_SOURCES = \
 	base64.h \
 	thread.h \
 	hash.c \
-	hash.h
+	hash.h \
+	tomltk.c \
+	tomltk.h
 
 TESTS = \
 	test_base64.t \
--- a/src/libutil/Makefile.am
+++ b/src/libutil/Makefile.am
@@ -27,7 +27,8 @@ libutil_la_SOURCES = \
 TESTS = \
 	test_base64.t \
 	test_hash.t \
-	test_tomltk.t
+	test_tomltk.t \
+	test_cf.t
 
 test_ldadd = \
 	$(top_builddir)/src/libutil/libutil.la \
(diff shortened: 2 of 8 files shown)


grondo:
This looks really useful to me. 

Some initial thoughts on the interface. Take these as just discussion thoughts, not any critical feedback yet

 * A better name for `cf_table_parse` might be `cf_table_check` or similar since parsing has already been done during `cf_table_update`.
 * In current form I'm not sure `cf_table_parse` would end up being useful, since the interface is set up such that every `cf_table_get_*` has to check return value anyway. An alternative would be to have all `cf_table_get_*` return the value of each get(), and are guaranteed to succeed if the key types were checked as present and valid during `cf_table_parse/check`. E.g. in later code I could pass the value of `cf_table_get_boolean` directly to a conditional instead of first setting the value to a variable. Otherwise, users of the cf class would probably have to copy all cf values to local variables in general.
 i.e.
```C
bool opt;
if (cf_table_get_boolean (cf, "myopt", &opt) < 0)
    die ("configuration error for config option myopt");
if (opt) {
   /* Do optional stuff */
}
```
vs
```C
if (cf_table_get_boolean (cf, "myopt")) {
  /* Do optional stuff */
}
 
 * This might be going too far, but should tables, arrays, and values all be special cases of a `cf_item` type (akin to `json_t`) so that you can get just `get()` an item then see if it casts to the type you wanted. This would be useful for arrays because you could get an item, make sure it is an array as expected with `cf_item_is_array()`, then iterate over it directly with `cf_item_next` or something. This would also allow arrays of arbitrary items be allowed (e.g. arrays of tables if that is possible with TOML)

Sorry for the hurried comments.

**Comment by garlick** [comment]
Hmm, thanks - that gives me some things to ponder!

**Comment by grondo** [comment]
probably stupid idea: since cf_table is implemented as a `json_t`, would it make any sense to just expose that interface directly to users? E.g. the "config" class could be used to read in one more TOML files and iteratively add them to a `json_t`, and would also offer convenience accessors for types that json doesn't support directly, like timestamp?

The jansson interface is already pretty nice for doing some of the things I was concerned about above, and would allow iterating over generic arrays, as well as the keys of a configuration "table".

Of course, the big drawback is that you make users of the config class dependent directly on the Jansson API, so it would be a chore if we ever tried to switch away from that approach.

Sorry if this idea is dumb.

**Comment by garlick** [comment]
I was thinking along the same lines.  I'm a bit reluctant to expose `json_t` for reasons we covered before - problems if a project wants to use a different JSON library for example, but maybe we can expose a similar interface, simplified.  Maybe something like this (for discussion):
```c
struct cf_table_create (void); // create empty table
void cf_container_destroy (struct cf_container *cf); // array or table

/* Parse TOML buf, updating table cf with results.
 */
int *cf_table_parse (struct cf_container *cf, const char *buf, int len,
                     char *errstr, int errstrsz);

/* Apply 'opts' to table 'cf' according to flags.
 */
int cf_table_check (struct cf_container *cf,
                    const struct cf_option opts[], int flags,
                    char *errstr, int errstrsz);

/* Retrieve a value from container (NULL if invalid)
 */
void *cf_get_in (struct cf_container *cf, const char *key); // table
void *cf_get_at (struct cf_container *cf, int index); // array

/* Translate a value to a scalar type
 * If val is NULL, these return 0, 0., "", false, and 0 respectively.
 */
int64_t cf_int64 (void *val);
double cf_double (void *val);
const char *cf_string (void *val);
bool cf_bool (void *val); 
time_t cf_timestamp (void *val);

/* Create a new container from value representing table or array
 */
struct cf_container *cf_table_create_from (void *val);
struct cf_container *cf_array_create_from (void *val);
```

So basically `struct cf_container` becomes the "object" that can be an array or table.

The idea would be that once you've passed `cf_table_check()` (as described before), then you could safely do something like 
```c
sleep (cf_in64 (cf_get_in (conf, "timeout")))
```
Getting warmer?

grondo:
> I'm a bit reluctant to expose json_t for reasons we covered before - problems if a project wants to use a different JSON library for example, but maybe we can expose a similar interface, simplified. Maybe something like this (for discussion):

Even if we hide the jansson interface behind another API, didn't we discover we could still run into a problem even loading two different JSON libraries into the same address space if they have conflicting symbols? (Not sure if this is germane, but if we go through the trouble to hide `json_t` in the interface just so others can use other json libs, only to have that still be impossible, it might be something to think about)

Either way, I like the idea of offering a simpler interface, so I'm pretty happy with the direction you are going.

> Getting warmer?

Yes, I think this is good. Looks very simple to use.

Just some more suggestions for simplification off the top of my head

 * I actually like Jansson's `json_error_t` struct because it boils down whether you want to get passed back error information into a single parameter. I'm not sure if `cf_table_check` will have source and line information, but it would be cool to separate that from the error string like `json_error_t` does as well.
 * Could `cf_get_in/at` return a `cf_container *` instead of `void *` or does that not fit the implementation? If you did it this way you don't seem to need `cf*create_from`, and you're not passing around `void *` (I guess this suggestion is *a lot* more like the `json_t` interface). If you went this route maybe the `cf_container` struct could just be a `cf_item` or something.
* Is `cf_table_parse()` to be renamed `cf_container_parse`? That seems a little bit awkward, maybe `update` works here, but what would happen if you try to "update" a container that is an array? Maybe `cf_container_parse/cf_container_update` would return an EINVAL error with `update not allowed on cf item that is not a table`

Sorry for the jumbled thoughts!

garlick:
I had to refresh my memory by rereading flux-framework/flux-core#853, but yes it does seem like there is still a hole to fall into in the case of json-c at least.

Hmm, pondering...

garlick:
Here's an attempt that basically makes cf_t an alias for json_t and wraps a subset of functions.

I think it cleans up most of your concerns.  The `json_error_t` style error details is a good one - let me have a look at that too.

```c
typedef void cf_t;

cf_t *cf_create (void);
void cf_destroy (cf_t *cf);
cf_t *cf_copy (cf_t *cf);
enum cf_type cf_typeof (cf_t *cf);

cf_t *cf_get_in (cf_t *cf, const char *key);
cf_t *cf_get_at (cf_t *cf, int index);

int64_t cf_int64 (cf_t *cf);
double cf_double (cf_t *cf);
const char *cf_string (cf_t *cf);
bool cf_bool (cf_t *cf);
time_t cf_timestamp (cf_t *cf);

int cf_array_size (cf_t *cf);

/* Update table 'cf' with info parsed from TOML 'buf'.
 * On success return 0.  On failure, return -1 with errno set.
 */
int cf_update (cf_t *cf, const char *buf, int len,
               char *errstr, int errstrsz);

/* Apply 'opts' to table 'cf' according to flags.
 * On success return 0.  On failure, return -1 with errno set.
 */
int cf_check (cf_t *cf,
              const struct cf_option opts[], int flags,
              char *errstr, int errstrsz);
```

garlick:
I added a `struct tomltk_error` to the tomltk class, and a `struct cf_error` to the cf class per your suggestion.  It just contains an error string right now but I'll try to expand it to be more comprehensive.

Per our offline discussion, I'll move forward on this iteration of the design and write some more tests.  The cf class should be considered a kind of "straw man" design.  If it doesn't pan out, I won't be broken hearted when we dispose of it.

grondo:
This sounds good, thanks!

garlick:
Finally got the coverage up and squashed.

grondo:
Looks like a call to your `cf_update_file` tests in `test_update_file()` is missing in `src/libutil/test/cf.c`

This will get the coverage even higher (and maybe poke codecov, which seems to be slumbering or something today)

garlick:
I fixed that, squashed, and repushed, but still no coverage.  It seems like it is aware of the update to the PR, but the coverage results are still the old ones (I think - at least cf_update_file is still red in the diff report)

https://codecov.io/gh/flux-framework/flux-security/pull/24

grondo:
Yeah, not sure if codecov is broken or slow today or what. I tested on my desktop and it looks great. Thanks!

garlick:
Thanks!

PR #25, opened 2017-12-13
libutil/cf: update header comments

garlick:
I really should have made a pass through the cf header, commenting each function.  This just adds that, no code changes.


grondo:
Sorry should have caught that in my review!

PR #26, opened 2017-12-15
add simple key=val encoding capability

garlick:
This PR adds a new `struct kv` class that implements a simple "key=value\0" serialization that should be easier to audit than JSON encoding.

In addition, the following changes were made to the `sigcert` class:
* move from `lib` to `libutil`, drop `flux_` namespace prefix
* internally represent "metadata" (arbitrary user data) using `struct kv`
* drop json certificate serialization, add base64 certificate serialization

### DIFF
--- a/src/libutil/Makefile.am
+++ b/src/libutil/Makefile.am
@@ -30,7 +30,8 @@ TESTS = \
 	test_base64.t \
 	test_hash.t \
 	test_tomltk.t \
-	test_cf.t
+	test_cf.t \
+	test_kv.t
 
 test_ldadd = \
 	$(top_builddir)/src/libutil/libutil.la \
@@ -62,3 +63,7 @@ test_tomltk_t_LDADD = $(test_ldadd)
 test_cf_t_SOURCES = test/cf.c
 test_cf_t_CPPFLAGS = $(test_cppflags)
 test_cf_t_LDADD = $(test_ldadd)
+
+test_kv_t_SOURCES = test/kv.c
+test_kv_t_LDADD = $(test_ldadd)
+test_kv_t_CPPFLAGS = $(test_cppflags)
--- a/src/libutil/Makefile.am
+++ b/src/libutil/Makefile.am
@@ -22,7 +22,9 @@ libutil_la_SOURCES = \
 	tomltk.c \
 	tomltk.h \
 	cf.c \
-	cf.h
+	cf.h \
+	kv.c \
+	kv.h
 
 TESTS = \
 	test_base64.t \
--- a/src/libutil/kv.c
+++ b/src/libutil/kv.c
@@ -1,0 +1,392 @@
+/*****************************************************************************\
+ *  Copyright (c) 2017 Lawrence Livermore National Security, LLC.  Produced at
+ *  the Lawrence Livermore National Laboratory (cf, AUTHORS, DISCLAIMER.LLNS).
+ *  LLNL-CODE-658032 All rights reserved.
+ *
+ *  This file is part of the Flux resource manager framework.
+ *  For details, see https://github.com/flux-framework.
+ *
+ *  This program is free software; you can redistribute it and/or modify it
+ *  under the terms of the GNU General Public License as published by the Free
+ *  Software Foundation; either version 2 of the license, or (at your option)
+ *  any later version.
+ *
+ *  Flux is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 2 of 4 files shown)


grondo:
Cool! This will be very useful. Is the kv class meant to replace hash.[ch] or used with it or just orthogonal?

Since kv encoding will be used as input to a process running with privileges, I wonder if we should experiment with fuzz testing of the decode interfaces? Here's an interesting fuzzer that I was looking at, though there are plenty of other ways to do it

http://caca.zoy.org/wiki/zzuf

dongahn:
zzuf sounds pretty interesting for debugging and testing in general. thanks you for the post @grondo!

garlick:
> This will be very useful. Is the kv class meant to replace hash.[ch] or used with it or just orthogonal

For most uses it would be orthogonal IMHO.  It doesn't currently include _delete_  or _update_ functions.  _get_ is an O(N) operation since it must do a linear search.   _put_ is also, since it first does a _get_ to check if a key already exists.  Values must be NULL-terminated strings, not pointers for example.

In short, more useful as a serialization format than a container object.

garlick:
As discussed I think I'll drop the sigcert changes from this PR and submit it separately.

grondo:
Two dumb ideas below.

> In short, more useful as a serialization format than a container object.

Hm, what if instead of a serialization format with a container interface, we had a container interface with added `serialize()` `deserialize()` methods? e.g. `char * hash_serialize (hash_t h)` and `hash_t hash_deserialize (hash_t h)`. Then the iteration, put/get/delete methods come with the container not the serializer? (Dumb idea?) `hash_t` might not be the right interface for this since there isn't a clean way to handle string keys that aren't also part of the value.

Also, it might be useful to have a binary serialization for kv -- for cases where we want to just send a `struct kv` directly over an fd or socket we can just write `len,buf` without unnecessarily going through base64 step since it is already an encoding. (Dumb idea?)

garlick:
Agreed, adding interfaces for binary serialization is probably the right move.

On  hash_t:  I was thinking that in critical security code, it would be advantageous to avoid the extra complexity of the hash, dynamic memory use, etc., but I do see your point.

Should we keep the limited container interfaces in struct kv for situations where we want to keep it simple (like reading security header when validating signatures), and add code (maybe a different module) to serialize a hash_t based hash-o-strings to struct kv and back again?  That might actually be more appropriate for the "meta" container in sigcert that direct use of struct kv...

grondo:
> Should we keep the limited container interfaces in struct kv for situations where we want to keep it simple (like reading security header when validating signatures), and add code (maybe a different module) to serialize a hash_t based hash-o-strings to struct kv and back again? That might actually be more appropriate for the "meta" container in sigcert that direct use of struct kv...

Yes, this sounds sensible

garlick:
OK, I added the "raw" encode/decode interface (binary serialization), fixed up coverage and squashed.  Mergeable?

grondo:
The one thing I'm not sure I like is the kv_buf_t type. However, I can see why you did it that way.
GNU `envz(3)` just bails on the key and returns an entry, leaving the caller to copy out the key, so I think the approach here is better.

It is impossible to update/delete a key? I can't think if that is requirement for now, but if you implemented `kv_delete()` then `kv_put()` could overwrite by calling `kv_delete()` first. (This is how `envz_add` works)

Besides those two minor questions I think this seems good to merge.

garlick:
Just remembered I wanted to do one more thing - change some params to const.

I could look at adding deletion and key update.  That would round out the "container" functionality...

Hmm, I could maybe replace `=` with `\0` as the key/value separator and then avoid kv_buf_t....

grondo:
> Hmm, I could maybe replace = with \0 as the key/value separator and then avoid kv_buf_t....

That's so crazy it just might work.

> I could look at adding deletion and key update. That would round out the "container" functionality...

I only mentioned it because it seemed strange to not have an interface to update a key (always get EEXIST with kv_put if the key exists). However, I can't think of a use case for that at the moment, so maybe we could delay addition of that feature until it is needed?

garlick:
Just pushed the const changes and a slight bit of refactoring I started for `kv_delete()`.  If it's OK I'd like to defer the other changes to another PR after I've had a bit more time to think about it (or until such time as they are needed).

garlick:
Just pushed a change to add `kv_delete()` and update `kv_put()` to allow existing keys to be updated.  

Tests added, and changes squashed.

OK, done for now!

PR #27, opened 2017-12-17
libutil/kv: increase size of kv_keybuf_t

garlick:
Just noticed an off by one error in the definition of kv_keybuf_t.  This fixes it.


garlick:
withdrawing this one - I've got another PR that eliminates the kv_keybuf_t as discussed.

PR #29, opened 2017-12-18
libutil/kv: eliminate kv_keybuf_t by changing serialization

garlick:
This PR changes serialization from "key=val\0key=val\0..." to "key\0val\0key\0val\0...".

This allows `kv_entry_key()` to return a key without needing to copy it to a user-provided buffer.

### DIFF
--- a/src/libutil/kv.c
+++ b/src/libutil/kv.c
@@ -117,7 +117,7 @@ static int kv_expand (struct kv *kv, int needsz)
 
 static bool valid_key (const char *key)
 {
-    if (!key || *key == '\0' || strlen (key) > KV_MAX_KEY || strchr (key, '='))
+    if (!key || *key == '\0')
         return false;
     return true;
 }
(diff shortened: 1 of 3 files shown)


grondo:
Nice simplification!

Wonder if you considered simplifying iteration since an "entry" is also the key?

garlick:
So like this maybe?
```c
const char *key;

key = kv_entry_first (kv);
while (key) {
    const char *val = kv_entry_val (key);
    ...
    key = kv_entry_next (kv, key);
}
```

grondo:
Yeah, I guess the `kv_entry_key` call is just redundant, but I was thinking you could also rename `entry` to `key` to make it clearer.

In fact if you treat `kv_next (kv, NULL)` the same as `kv_first()` then you could drop that call as well and have

```C
const char *key = NULL;
while ((key = kv_next (kv, key))) {
    const char *val = kv_val (key);
}
```

However, that might be going too far, and in retrospect I'm fine with the current interface.

garlick:
Or mabye
```c
const char *key = NULL;
const char *val;

while (kv_foreach (kv, &key, &val)) {
    ...
}
```

garlick:
Oops sorry, didn't see your other suggestion..

grondo:
> In fact if you treat kv_next (kv, NULL) the same as kv_first() then you could drop that call as well and have

This is how `argz_next()` works (if argument is NULL it returns the first entry) so there is actually precedent in common usage for that interface...

garlick:
The thing that gives me pause is that `kv_next()` doesn't work on just any key; it needs to be one returned from `kv_next()` (or NULL).  Hmm...

grondo:
Is that different from the current way kv_entry_next works? I apologise if
so

garlick:
Sorry, slow to get my head around it.  I think it's OK, let me try it.

grondo:
> The thing that gives me pause is that kv_next() doesn't work on just any key; it needs to be one returned from kv_next() (or NULL). Hmm...

I see what you are saying. That might mean your existing "entry" terminology is more clear.

garlick:
OK, I went with `kv_next()` and `kv_val()` as in your description above.  It works for me if it works for you.

grondo:
Nice, looks good to me! Merge?

garlick:
Yes please.

PR #30, opened 2017-12-19
libutil/kv: add typed values and split/join capability

garlick:
Prepend a "type designator" byte to kv values so that string, int64_t, double, bool, and timestamp typed values can be supported.   Add typed get, put, and iteration value access functions.

Add `kv_join()` to add entries from one kv object to another, optionally prepending a key prefix; and add `kv_split()` to do the reverse.

These two relatively simple additions allow typed, hierarchical data such as TOML or JSON to be serialized as a flat kv buffer, and extracted again.

### DIFF
--- a/src/libutil/kv.c
+++ b/src/libutil/kv.c
@@ -32,7 +32,6 @@
 #include <stdarg.h>
 #include <assert.h>
 
-#include "base64.h"
 #include "kv.h"
 
 #define KV_CHUNK 4096
@@ -41,7 +40,6 @@ struct kv {
     char *buf;
     int bufsz;
     int len;
-    char *base64;
 };
 
 void kv_destroy (struct kv *kv)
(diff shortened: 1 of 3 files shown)


grondo:
Is it an error from kv_join() if kv1 and kv2 contain the same key when no prefix is used, or is it an overwrite? I'm guessing that kv_join() won't be used often without a prefix, but probably the behavior either way should be documented in the header (and checked in tests).

`envz_merge()` has a neat feature where an "unset" key in the envz object being merged will result in the key being unset in the result. I can't think of why we'd need that for kv, but something to keep in mind as a use case for "null" values in the future if needed.

garlick:
Thanks!  I documented the overwrite behavior in the header and added a test for it.

I avoided representing NULL values so far but we could easily add that if a need develops.

garlick:
Squashed.

grondo:
Cool, I'll merge when checks complete.

PR #32, opened 2017-12-20
libutil/kv: add varargs get/put

garlick:
@grondo - what do you think about this change to the kv api which replaces the typed kv_get_TYPE, kv_put_TYPE functions with single varargs functions?

This seemed like a good idea until I recalled the fact that one can get into trouble with the lack of type promotion through varargs interfaces, specifically
```c
kv_put (kv, key, KV_INT64, 42);
```
causes an out of bounds memory access. You need
```c
kv_put (kv, key, KV_INT64, 42LL);
```

Maybe not such a good idea in a security sensitive situation?

OTOH, the reduction in API footprint is nice, especially given that this class may end up partially wrapped in several places.

Hope you don't mind me posting this half baked idea just to get your feedback.

### DIFF
--- a/src/libutil/kv.c
+++ b/src/libutil/kv.c
@@ -33,6 +33,7 @@
 #include <inttypes.h>
 #include <stdbool.h>
 #include <time.h>
+#include <stdarg.h>
 #include <assert.h>
 
 #include "kv.h"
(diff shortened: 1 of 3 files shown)


grondo:
Yeah, the number of API calls would be reduced, but actually calling them seems a little more cumbersome and error prone, so I'm not sure it is a complete win? I do like the idea though...

Are the kv_get() varargs calls pretty safe though? You wouldn't want to end up in a situation where a kv buffer could be manipulated in such a way as to cause invalid memory read or write on the reader side, but does that seem fairly safe?

garlick:
> Are the kv_get() varargs calls pretty safe though? You wouldn't want to end up in a situation where a kv buffer could be manipulated in such a way as to cause invalid memory read or write on the reader side, but does that seem fairly safe?

I think it is safe in that respect.  There is only the KV_INT64 integer type, so all integers look the same "on the wire".  A `kv_get` call fails if the input value has a different type than the one requested.

Calling wise it's:
```c
int64_t i = 42;
if (kv_put_int64 (kv, key,  i) < 0)
    goto error;
if (kv_get_int64 (kv, key, &i) < 0)
    goto error;
```
versus
```c
int64_t i = 42;
if (kv_put (kv, key, KV_INT64, i) < 0)
    goto error;
if (kv_get (kv, key, KV_INT64, &i) < 0)
    goto error;
```
so IMHO about the same calling cumbersomeness.  The issue for me is the lack of help you get from the compiler if the argument types are wrong.

grondo:
> so IMHO about the same calling cumbersomeness.

I counted at least 3 more characters per call, and uppercase! :stuck_out_tongue: 

I wonder if there is some __attribute__ foo that can help the compiler check type safety? Probably not, but there is also this fun approach (I'm considering the struct/designated initializer approach)

https://snai.pe/c/preprocessor/varargs/

grondo:
I'm actually fine with this approach if the `get` call is safe. I like the simplified interface, if type safety is really needed then maybe we can try some hackery as in above or the calls can just be wrapped.

garlick:
OK, I'll add a note about the default integer argument promotion not matching the expected int64_t type for kv_put (KV_INT64).  That seems to be the easiest pothole to fall into here.

grondo:
Ready to merge? Looks fine to me!

garlick:
Sure, thanks.

PR #33, opened 2017-12-20
libutil/cf: add interface to update a cf table with multiple files

grondo:
This PR adds a `cf_update_glob()` call to iteratively call `cf_update_file()` on all files matching a pattern. It is meant to be used for loading TOML files from a `conf.d` style directory, e.g. `cf_update_glob (cf, "/etc/conf.d/*.toml")`.

Some decisions here were made that probably need to be vetted. The call only updates the `cf` table if *all* files matching the pattern parse and are loaded successfully. That seemed the most sane way to do it.

The return value from the function is the number of files successfully parsed, or -1 on an unrecoverable error (ENOMEM or if `glob(3)` can't open a directory). A return value of zero therefore means there was no error, but no files matched the pattern. I figured this was not necessarily an error condition, but the caller may want to test for it.

### DIFF
--- a/src/libutil/cf.c
+++ b/src/libutil/cf.c
@@ -27,6 +27,7 @@
 #endif /* HAVE_CONFIG_H */
 #include <errno.h>
 #include <string.h>
+#include <glob.h>
 #include <jansson.h>
 
 #include "src/libtomlc99/toml.h"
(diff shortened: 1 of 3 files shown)


grondo:
codecov is failing because of error conditions.

garlick:
Thanks!  Also, thanks for fixing the cut-and-paste-o and tightening up errprintf().

I think your choices on the return value, and the "all or nothing" update semantics are spot on.

I'm happy to merge this if you're ready.

Issue #19, opened 2017-11-25
implement signed public key certificates


garlick:
We have discussed (offline) signing user public certificates with an IMP private key.

A signed request _J_ could include the user public certificate.  Validation of _J_ on the receiving end would only require access to the IMP/CA public key to validate the user certificate, then the user certificate could be used to verify the signature.  An advantage of this approach is that the validator doesn't need a "keyring" or similar of all acceptable user public keys.  It would just need the CA public key and a revocation list.

I didn't realize it but OpenSSH has a similar facility.  Here's a description:

https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/deployment_guide/sec-using_openssh_certificate_authentication

grondo:
Now that you mention it, I remember reading this Facebook post awhile ago on this subject:

https://code.facebook.com/posts/365787980419535/scalable-and-secure-access-with-ssh/

Sorry, I had forgotten about it

PR #35, opened 2017-12-28
libutil/signer:  add class for signing/verifying arbitrary payloads

garlick:
This is a first cut at a class for signing IMP input (or other things).

It depends on a config file that contains something like
```toml
[signer]
ttl = 3600
ca-cert-path = "/etc/flux-security/ca-cert"
```

It hard-wires a simple certificate authority scheme, where user certs are signed by a CA cert.  The user cert is then enclosed with the signed material.  The verifier checks the signature with the enclosed public key, then checks the userid claim in the public key with the CA public key.

We'll want to make the mechanism configurable but I felt we need to get one working first.

There's more detail in the description of commit  0285f9f

Testing is just the bare minimum at this point.  I wanted to get this up for any feedback though.


grondo:
This looks nice! 

The one thing I'm not sure about is passing in the signer "configuration" as a config object. Perhaps it would be a better interface to pass in the ca-cert-path or the cert itself into signer_create() and the ttl for each signature or as a setopt type interface? This would allow more flexibility in how signers are configured (i.e. it may not make sense for a certain entity to have a `[signer]` table) and also describe in the interface exactly which options are being set instead of hiding them behind a `cf_t`.

That being said, there are some benefits for configuration via `cf_t`, so perhaps those outweigh any misgivings.

garlick:
My thought in doing it that way was that different signing mechanisms might require a different configuration "schema" (e.g. ca-cert-path is specific to this particular scheme of signing user curve keys with a CA), and the user of the signing interface might not want to have to dig out mechanism-specific config values in order to pass them in via setopt or whatever.

I'm not too sure I understand where the boundaries should be drawn between general and mechanism specific code.  In this iteration it's all lumped together.  I thought maybe it would become clearer as we get further in...

I'm not necessarily disagreeing with the suggestion of a setopt interface instead of a cf_t - just talking through my thought process.

grondo:
> My thought in doing it that way was that different signing mechanisms might require a different configuration "schema" (e.g. ca-cert-path is specific to this particular scheme of signing user curve keys with a CA), and the user of the signing interface might not want to have to dig out mechanism-specific config values in order to pass them in via setopt or whatever.

This is a good point. I'm not sure I see the path to multiple signing mechanisms so I'm with you on not seeing the boundaries yet. In this scheme how would the other mechanisms be activated? Would it be some magic in the configuration `cf_t`? 

Sorry just thinking out loud here, and like you said it will be easy to change the interface in the future if this isn't quite right.

garlick:
Continuing with the thinking out loud, maybe the `[signer]` table would contain a key that selects the mechanism type, and then a sub-table for it, like
```toml
[signer]
mech = "flux-ca"
[signer.flux-ca]
ca-cert-path = "..."
ttl = 3600
```

grondo:
I guess what also bothers me about the use of the configuration table to initialize the signer is that you can never have a signer with parameters that could possibly be overridden on the cmdline. To programmatically change anything, you'd have to write a string config in TOML and call `cf_update`. Maybe this is actually a feature, but I'm not quite there yet.

Also, I'm not sure but does this approach require that the configuration tables for all entities using this interface have a `signer` table at the top level? (e.g. the imp couldn't have `[imp.signer]` and pass this down to signer create?) Maybe that could be solved by not requiring a `[signer]` table to be passed to `signer_create`, but instead a `cf_t` that contains an appropriate `[signer]` table with required contents? Then the creator could pull the appropriate table out of whatever config it has loaded, without placing restrictions on the signers other configuration table layout?

garlick:
> I guess what also bothers me about the use of the configuration table to initialize the signer is that you can never have a signer with parameters that could possibly be overridden on the cmdline. To programmatically change anything, you'd have to write a string config in TOML and call cf_update. 

Well that's a good point.  Would it be best to have a setopt interface in the cf class?

> Maybe that could be solved by not requiring a [signer] table to be passed to signer_create, but instead a cf_t that contains an appropriate [signer] table with required contents? Then the creator could pull the appropriate table out of whatever config it has loaded, without placing restrictions on the signers other configuration table layout?

I was having the same thought.  I'll make this change.

grondo:
> Well that's a good point. Would it be best to have a setopt interface in the cf class?

I thought of that, but it felt a bit wrong. However, perhaps that would indeed be useful, because users of the class could first set defaults, then load configuration and update defaults with `cf_update()` (or override with environment variables, cmdline options).

garlick:
Just forced an update after some refactoring.

I pulled out a `ca` class that implements certificate sign, verify, and revoke.  The configuration is "owned" by this class now, and behaves as discussed (the ca table is passed in, not its parent table).
The ca class is described in detail in its commit message 285d8bd82b2b20a81a1c00a9b6e4f3f42628d68d and in [ca.h](https://github.com/garlick/flux-security/blob/285d8bd82b2b20a81a1c00a9b6e4f3f42628d68d/src/libutil/ca.h)

The `signer` class then uses the `ca` class during verification.  The `ca` class is not involved in signing.
The signer class is described in detail in its commit message 886a2ae17689124eb7442a3cbc67924ab44b751e  and in [signer.h](https://github.com/garlick/flux-security/blob/886a2ae17689124eb7442a3cbc67924ab44b751e/src/libutil/signer.h)

Tests are still extremely basic.

grondo:
The abstractions here do seem a lot clearer now. I'm still getting hung up on the nomenclature here I think, though. To me it is still a little confusing to create instances of "signer" and "CA" classes just to validate a signature. In the case of the `ca` class, if creating a ca only for validation, are the `cf_t` configuration parameters required and ignored, silently ignored, or used somehow?

I'm also assuming the cert-path is to a directory in which the private part of the key would be allowed to be missing?

It feels like I want a "signer" and a "verifier"/"validator" not just a signer and ca, but I'm probably just not understanding the problem space. I apologize.

garlick:
> To me it is still a little confusing to create instances of "signer" and "CA" classes just to validate a signature. In the case of the ca class, if creating a ca only for validation, are the cf_t configuration parameters required and ignored, silently ignored, or used somehow?

Part of validating a signature in this scheme is validating the signed cert enclosed within the security header.  That requires access to the CA public key and the CA revocation list.  The CA configuration points to those things in the file system.  That is why the ca object is passed into the signer as part of verification.

> I'm also assuming the cert-path is to a directory in which the private part of the key would be allowed to be missing?

It uses the sigcert semantics for that (inherited from zcert).  The specified path is to the secret key, and the public key is that plus `.pub`.   It turns out to be kind of convenient that way.

> It feels like I want a "signer" and a "verifier"/"validator" not just a signer and ca

Well the bulk of what is going on in that class is "wrapping" and "unwrapping" serialized signed payload.  Since these are inverse activities it kind of makes sense to have them in the same class.  Maybe a class rename is in order?  I'm kind of stumped thinking up a better name that encompasses both signing and verifying (or wrapping and unwrapping) right now though...

garlick:
I should have referenced issue #19 in here somewhere...

garlick:
I think this is in a pretty good state for a beginning.  The code coverage could be a bit better, but I think I've reached the point where it would be best to move forward and find problems with the high level design, rather than spend another day writing the more involved tests.

I had developed a "imp ca-sign" example, and modified the keygen example to work with this, and that did inform the design a bit, but ultimately I scrapped those since they probably served their purpose and would be doing no good checked into the code base with no users.

grondo:
Ok, merging this so we can move forward. Sorry @garlick I lost track of it earlier!

PR #40, opened 2018-01-20
Add basic skeleton for the IMP

grondo:
This PR adds basic functional skeleton for the IMP including:
 
 * Support for directory of TOML config files under a directory like `imp.conf.d/*.toml`. For now config is only loaded from build directory, but support is there to load from hard-coded system configuration directory by building a separate binary (e.g. flux-imp.installed or something).
 * Implementation of privilege separation when IMP is running setuid. The IMP process will fork off a child that drops privileges to the real UID. The "privliged parent" and "unprivileged child" are connected with a pair of pipes. The child is meant to process all input from the untrusted source, and push data back to the parent over the pipe in a more restricted format. Nothing uses this yet.
 * Support for running under sudo by simulating a setuid installation with sudo (set real UID to SUDO_USER)
 * Very basic subcommand support. Subcommands are split into an "unprivileged" child half and "privileged" parent half. Only a child subcommand may be registered if the command does not require privileged.
 * Addition of a simple "version" subcommand which does not require privilege to run.

Additionally, a `t` directory is added with a sharness support. The sharness suite here uses a hook from MUNGE to detect non-interactive sudo support, and uses that to run unit tests and modes of `flux-imp` that require privilege. Not sure if this is the best approach.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,7 @@ before_install:
   - export PKG_CONFIG_PATH=$HOME/local/lib/pkgconfig:$HOME/local/share/pkgconfig
 
 script:
+  - export FLUX_TESTS_LOGFILE=t
   # Force update to allow version m4 to work
   - ( git fetch --unshallow --tags || true )
   - ./autogen.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,8 @@ before_install:
   - export PKG_CONFIG_PATH=$HOME/local/lib/pkgconfig:$HOME/local/share/pkgconfig
 
 script:
+  # Force update to allow version m4 to work
+  - ( git fetch --unshallow --tags || true )
   - ./autogen.sh
   - ./configure ${ARGS}
   - eval ${MAKECMDS:-make && make check}
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,3 +38,6 @@ script:
 
 after_success:
   - if test "$COVERAGE" = "t"; then bash <(curl -s https://codecov.io/bash); fi
+
+after_failure:
+  - find . -name t[0-9]*.output | xargs -i sh -c 'printf "\033[31mFound {}\033[39m\n";cat {}'
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: c
-sudo: false
+sudo: true
 
 matrix:
   include:
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,4 +1,4 @@
-SUBDIRS =         src
+SUBDIRS =         src t
 ACLOCAL_AMFLAGS = -I config
 EXTRA_DIST = \
 	config/tap-driver.sh \
--- a/configure.ac
+++ b/configure.ac
@@ -80,6 +80,7 @@ AX_CODE_COVERAGE
 #
 AC_CONFIG_FILES( \
   Makefile \
+  t/Makefile \
   src/Makefile \
   src/lib/Makefile \
   src/libtap/Makefile \
--- a/src/imp/Makefile.am
+++ b/src/imp/Makefile.am
@@ -22,7 +22,9 @@ flux_imp_SOURCES = \
 	imp_log.c
 
 TESTS = \
-	test_imp_log.t
+	test_imp_log.t \
+	test_privsep.t
+
 check_PROGRAMS = \
 	$(TESTS)
 
@@ -40,3 +42,12 @@ test_imp_log_t_SOURCES =  \
 	imp_log.h
 
 test_imp_log_t_LDADD = $(test_ldadd)
+
+test_privsep_t_SOURCES = \
+	test/privsep.c \
+	privsep.c \
+	privsep.h \
+	imp_log.h \
+	imp_log.c
+
+test_privsep_t_LDADD = $(test_ldadd)
--- a/src/imp/Makefile.am
+++ b/src/imp/Makefile.am
@@ -47,6 +47,8 @@ test_privsep_t_SOURCES = \
 	test/privsep.c \
 	privsep.c \
 	privsep.h \
+	sudosim.h \
+	sudosim.c \
 	imp_log.h \
 	imp_log.c
 
--- a/src/imp/Makefile.am
+++ b/src/imp/Makefile.am
@@ -22,6 +22,9 @@ flux_imp_SOURCES = \
 	imp_log.h \
 	imp_log.c
 
+EXTRA_DIST = \
+	imp.conf.d
+
 TESTS = \
 	test_imp_log.t \
 	test_privsep.t \
--- a/src/imp/Makefile.am
+++ b/src/imp/Makefile.am
@@ -23,7 +23,8 @@ flux_imp_SOURCES = \
 
 TESTS = \
 	test_imp_log.t \
-	test_privsep.t
+	test_privsep.t \
+	test_impcmd.t
 
 check_PROGRAMS = \
 	$(TESTS)
@@ -53,3 +54,10 @@ test_privsep_t_SOURCES = \
 	imp_log.c
 
 test_privsep_t_LDADD = $(test_ldadd)
+
+test_impcmd_t_SOURCES =  \
+	test/impcmd.c \
+	impcmd.c \
+	impcmd.h
+
+test_impcmd_t_LDADD = $(test_ldadd)
--- a/src/imp/Makefile.am
+++ b/src/imp/Makefile.am
@@ -22,6 +22,12 @@ flux_imp_SOURCES = \
 	imp_state.h \
 	imp_log.h \
 	imp_log.c \
+	privsep.c \
+	privsep.h \
+	impcmd-list.c \
+	impcmd.c \
+	impcmd.h \
+	version.c \
 	testconfig.c
 
 testconfig.c: $(top_builddir)/config/config.h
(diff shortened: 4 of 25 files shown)


grondo:
Hm, non-interactive sudo isn't working in travis for this PR as it was in my personal repo tests. 

Before merging allow me to figure that one out...

grondo:
Ok, I was confused and `sudo: false` was still set in `.travis.yaml`. Fixed and now sudo-based tests are running. Coverage is still a low, but I wanted to submit this PR to get early feedback before figuring out exactly how security sensitive testing under privilege is going to work.

garlick:
Ah great!  I'll try to allocate some time to give this a review over the weekend.

garlick:
To be sure I understand, if the imp is installed setuid, and `allow-sudo = true`, would I then be able to do something like
```
$ SUDO_USER=grondo imp ca-sign
```
to sign a credential as you, since the imp would see the real uid as grondo?

grondo:
No, I don't think that will work because the *real* UID is not root, it would be your UID. Sudo support should only be active if real uid == effective uid == 0.

However, it would be a great idea to explicitly defend against this, and have the IMP log a security message or audit log when it is attempted. And of course a test (but I don't have a great way to test the setuid mode right now. Perhaps we need to `make install` and run a suite of tests against the installed version of this stuff. More work needed here)

garlick:
Oh sorry, I didn't spot where real == effective == 0 was being checked.

grondo:
> Oh sorry, I didn't spot where real == effective == 0 was being checked.

The "sudosim" code only runs if `geteuid() == 0` and `sudo_user_name()` only checks `SUDO_USER` when `getuid() == 0`.

It wouldn't hurt if the code was more explicit though, and specifically checked for the case where the real UID and SUDO_USER are different so that an error or audit log could be issued.

garlick:
This looks very nice.  Some minor comments:
* as discussed offline, the cached uid/gid values aren't currently used (which is fine by me, though you indicated it might be better to leave them out until there is a user)
* looks like `imp_cmd_find_parent()` is not called, so the privileged component of a command would not execute, if provided?
* would it make sense to implement a "selftest" subcommand that checks that the two halves of a command have the expected credentials in the context of the main IMP executable (including when installed?)

grondo:
Good points. I'll clean up the cached uid/gid values and we can add them only when/if they are needed.

> looks like imp_cmd_find_parent() is not called, so the privileged component of a command would not execute, if provided?

Yeah, this code is unfinished. The way it  is designed to work is that the unprivileged code will read argv for the subcommand string (really just `argv[1]` I suppose), and then process IMP input using the unprivileged child function. The input would then be massaged into simple serialization format suitable for transmitting up to the privilged parent. The "command" would be prepended or otherwise be a part of the data transmitted to the parent, such that the first thing the parent reads off the line is the "command", and then the rest of the serialized input would be processed by the privileged component of the function.

This approach delays activation of any privileged code until latest point possible. But since there isn't an implemented command that requires privilege yet, that code is not exercised.

>  would it make sense to implement a "selftest" subcommand that checks that the two halves of a command have the expected credentials in the context of the main IMP executable (including when installed?)

I was having the same thought. And this will be a good first implementation of a command requiring privilege and thus will satisfy your request above to at least implement one "privlieged parent" function.

garlick:
Ah, that makes sense!  Well, if you prefer to get this merged and then add that in a follow-on PR, I'm good with that.

grondo:
No, I think the selfcheck or `flux-imp whoami` command makes great sense and will give us more confidence that the privilege separation is working correctly before merging this PR.

grondo:
Really good suggestion to develop at least one privileged IMP command as part of this PR. Trying to write `flux-imp whoami` uncovered some awkwardness with the current interface and I'm reworking it a bit now.

grondo:
Ok, I pushed some experimental changes here, including:

 * Split the imp command table into its own file `impcmd-list.c` so that  `test/impcmd.c` doesn't have to be linked with every future command.
 * Restructure how IMP "commands" work and use `struct kv *` directly to reduce boilerplate code. The child command callback will be passed a `struct kv *` with cmd=argv[1] already preset. Once child command finishes parsing/marshalling all other input, it can fill in the rest of the requisite values and pass to the privileged parent. Privileged parent reads struct kv and passes to the parent command.
 * Add a simple `flux-imp whoami` command to print UID/GIDs in the context of each child/parent commands.

```
$ sudo ./flux-imp whoami
flux-imp: unprivileged: uid=6885 euid=6885 gid=6885 egid=6885
flux-imp: privileged: uid=6885 euid=0 gid=6885 egid=0
```

Still todo: the privsep code should probably export functions that write/read `struct kv` instead of the low level read/write functions. The privsep protocol should be rethought to allow sending of more than one `struct kv` back and forth between parent/child (though that isn't required right  now)

grondo:
Ok, a bit of rework here and rebasing to clean up history. Main change since the last iteration is a (slightly) better method for sending `struct kv` back and forth between unprivileged child and parent using the `privsep_write_kv()` and `privsep_read_kv()` functions added to the privsep interface. 

For now a simple approach is taken, and each `struct kv` is prefixed with the incoming length, to allow the potential for more than a single kv to be sent, though the current implementation doesn't require it.

Still todo:
 * Protect `privsep_read_kv()` to reject incoming `struct kv` that are outside of reasonable size bounds (not sure what that should be, but I'll likely just hard-code something for now). This will protect against attacker causing the privilged code to request a very large amount of memory.
 * Still need to add some checks for malicious behavior (e.g. real uid != SUDO_USER, etc), and log a security message.

grondo:
To summarize, the simplistic approach for adding a new IMP command is to write an "unprivileged" (child) function with the prototype:
```C
int unprivilged_cmd (struct imp_state *imp, struct kv *kv);
```
The `struct kv` passed into the function has been initialized with the key/value `cmd="argv[1]"`. This function should then do as much as possible (gather input, interact with filesystem as user, etc) while unprivileged, and insert data required for the *privileged* work into `kv` as new key/value pairs. Once the unprivileged work requires help from the privileged parent, the `kv` is written to the parent with a call like
```C
privsep_write_kv (imp->ps, kv);
```
Once the first `struct kv` is received on the privilege separation pipe, the parent function corresponding to the value of the `"cmd"` key  in  the received `struct kv` is invoked, and that `kv` is passed to the parent function using the same prototype as the unprivileged child above. The parent pulls out necessary keys from `kv` and does the privileged work.

If there is no corresponding parent command then the privileged parent does nothing (which is reasonable for IMP command that requires no privilege, e.g. "version").

If the IMP is run unprivileged (e.g. not under sudo nor setuid), then the privilege separation handle `imp->ps` is NULL. Child functions should check for this condition and either issue an error, silently exit, or run in some sort of "unprivleged mode", functionality depending.

At this point, new commands should be manually added to the impcmd_list array in `impcmd-list.c`

grondo:
Ok, pushed some protection for insane size of kvs written to/from the privsep pipe, and tests of this functionality. 

Last thing I was going to do here is generate an error for the case @garlick noted above, where `SUDO_USER` is set in the environment but real uid is not zero and doesn't match `SUDO_USER`. (Currently this condition is silently ignored and the IMP will refuse to run as root). Eventually this error should generate an audit or security event, for now probably just a fatal error.

garlick:
Does the "giant" input test need a corresponding "runt" test?

grondo:
> Does the "giant" input test need a corresponding "runt" test?

Yes, good point. A "size" of zero or negative integers or other invalid input should be attempted. (Actually eventually I'd like to simulate the "parent" end of the privsep pipe as a process that takes arbitrary input on stdin and just fuzz the thing or run through a series of invalid and valid input)

grondo:
Ok, protection in sudo code activated but I need to add tests.

That, and expanding the "invalid" kv input tests for the privsvep code and this PR should be close to ready for full review.

grondo:
> Ok, protection in sudo code activated but I need to add tests.

Nevermind, this isn't working because the sudosim code is only ever invoked if `sudo_is_active()` returns true, which means the real UID is 0 and SUDO_USER is set, therefore none of the extra paranoid checks are even valid.

If it is OK, the original approach of just ignoring `SUDO_USER` for any real UID except root seems the most straightforward. Though then we can't issue an auditable event or log message when someone tries to set `SUDO_USER` from the command line before running the setuid IMP, it doesn't seem like that is necessary (and it becomes awkward) when that variable is being ignored anyway.

Later, the IMP could have a configurable list of env variables to sanitize when running setuid and SUDO_USER could explicitly be one of those.

Let me know your thoughts.

garlick:
That makes sense - works for me.

grondo:
Ok, I've reverted the ineffective change to the sudosim code and expanded the privsep_read_kv() tests to check for more out of range values.

I also expanded the IMP sharness tests with a few more sanity checks like SUDO_USER is ignored if UID != 0, invalid SUDO_USER causes error, etc. Also, if sudo is available in the test, `flux-imp` is copied and mode is changed to setuid to get some sanity tests for setuid installed IMP.

It is probably going to be difficult to get much more coverage that what we have here, though now that there is a way to test setuid version perhaps I can eke out a few more boundary case tests.

garlick:
Test coverage looks pretty good.  I'm OK with merging now if you are.

grondo:
Let me squash the latest two commits and then I'm fine with merging...

grondo:
Ok , I squashed all the privsep incremental improvements. Give it a quick review but this is probably ok to merge now if tests pass.

garlick:
Completed another spin through and didn't spot anything wrong, so I'll merge when travis passes.
Looking forward to using the privsep interface - it is pretty cool!

grondo:
Had to restart a couple builders that failed apt-get.

grondo:
Yay, thanks!

Issue #42, opened 2018-02-02
imp subcommand for CA signing of public certs


garlick:
User signing certs must be signed by the CA private key.  As discussed in #41, the `flux keygen` utility would use a flux security API to generate the user certs and get the public cert signed by the CA.

The API needs to call out to an IMP sub-command that can access the CA private key.  Input would be the unsigned cert and other options (TBD).  Output would be the signed cert.

garlick:
Closing after merge of #48

PR #48, opened 2018-02-07
imp/casign: add casign subcommand

garlick:
I thought I'd put this up for comment (mainly to see if I'm using privilege separation correctly).

This adds a `casign` subcommand that accepts an unsigned user cert on stdin, signs it with the CA cert, and emits the signed user cert on stdout.  It's a very tentative stab at implementing #42.

It would be called by a flux-core keygen utility as part of generating a user cert.

The unpriv child just reads the cert in encoded kv form on stdin and adds it to the kv that is sent to the priv parent.  The parent uses its privilege to access the CA cert and uses it to sign the user cert (as the real uid), then emits the signed cert in encoded kv form on stdout.

No TOML config file, no source tree cert location, and no tests yet.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -2,6 +2,7 @@ coverage:
  round: down
  range: "50...100"
  ignore:
+ - "t/.*"
  - ".*/test/.*"
  - ".*/libtap/.*"
 
--- a/src/imp/Makefile.am
+++ b/src/imp/Makefile.am
@@ -32,7 +32,8 @@ flux_imp_SOURCES = \
 	version.c \
 	whoami.c \
 	testconfig.h \
-	testconfig.c
+	testconfig.c \
+	casign.c
 
 testconfig.o: testconfig.h
 testconfig.h: $(top_builddir)/config/config.h
--- a/src/imp/casign.c
+++ b/src/imp/casign.c
@@ -1,0 +1,152 @@
+/*****************************************************************************\
+ *  Copyright (c) 2017 Lawrence Livermore National Security, LLC.  Produced at
+ *  the Lawrence Livermore National Laboratory (cf, AUTHORS, DISCLAIMER.LLNS).
+ *  LLNL-CODE-658032 All rights reserved.
+ *
+ *  This file is part of the Flux resource manager framework.
+ *  For details, see https://github.com/flux-framework.
+ *
+ *  This program is free software; you can redistribute it and/or modify it
+ *  under the terms of the GNU General Public License as published by the Free
+ *  Software Foundation; either version 2 of the license, or (at your option)
+ *  any later version.
+ *
+ *  Flux is distributed in the hope that it will be useful, but WITHOUT
+ *  ANY WARRANTY; without even the IMPLIED WARRANTY OF MERCHANTABILITY or
+ *  FITNESS FOR A PARTICULAR PURPOSE.  See the terms and conditions of the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License along
+ *  with this program; if not, write to the Free Software Foundation, Inc.,
+ *  59 Temple Place, Suite 330, Boston, MA 02111-1307 USA.
+ *  See also:  http://www.gnu.org/licenses/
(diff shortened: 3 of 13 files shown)


garlick:
Per offline discussion, I'll rework this to take TOML on stdin and stdout so we don't have to export the kv class.

garlick:
I think at this point the coverage is OK and this is ready for a review.

By the way, in this test, a `ca.toml` file is added to the imp source directory, and the sharness test for the casign subcommand assumes this config is in place.  Later, as discussed offline, we might want to make the config path configurable for in-tree testing.   I'll revisit these tests then and have the sharness script generate a config specific to the tests.

garlick:
Er, actually I should squash down one commit, and fix a buglet in `certutil.c`.  Sorry...

garlick:
Sure, let's wait and avoid the extra churn.

garlick:
Rebased on #50, with `ca.toml` now generated in the test directory.  Thanks @grondo!

garlick:
Rebased after merge of #50 and fixed my sudo test as discussed over there.

grondo:
Tested out this PR on my local machine with some "bad input" and perhaps `sigcert_fread_*` needs some kind of defensive coding, if possible

```
$ src/imp/flux-imp casign < /dev/zero
Segmentation fault (core dumped)
```

Since the read is occurring in unprivileged process, this isn't a critical issue, but it would be nicer to throw an error in this case.

garlick:
Uh oh!  Yes, will add a test and get that sorted.

garlick:
backtrace from core file
```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  __memcpy_sse2 () at ../sysdeps/x86_64/multiarch/../memcpy.S:201
201	../sysdeps/x86_64/multiarch/../memcpy.S: No such file or directory.
(gdb) bt
#0  __memcpy_sse2 () at ../sysdeps/x86_64/multiarch/../memcpy.S:201
#1  0x00007fcbf550dfd3 in __GI__IO_file_xsgetn (
    fp=0x7fcbf58598e0 <_IO_2_1_stdin_>, data=<optimized out>, n=1000)
    at fileops.c:1383
#2  0x00007fcbf5503236 in __GI__IO_fread (buf=buf@entry=0x7fca87c96170, 
    size=size@entry=1, count=count@entry=1000, 
    fp=fp@entry=0x7fcbf58598e0 <_IO_2_1_stdin_>) at iofread.c:38
#3  0x000000000040cd37 in fread (__stream=0x7fcbf58598e0 <_IO_2_1_stdin_>, 
    __n=1000, __size=1, __ptr=0x7fca87c96170)
    at /usr/include/x86_64-linux-gnu/bits/stdio2.h:295
#4  toml_parse_file (fp=fp@entry=0x7fcbf58598e0 <_IO_2_1_stdin_>, 
    errbuf=errbuf@entry=0x7ffde82a5950 "", errbufsz=errbufsz@entry=200)
    at toml.c:1306
#5  0x0000000000406e2d in sigcert_fread_public (
    fp=0x7fcbf58598e0 <_IO_2_1_stdin_>) at sigcert.c:599
#6  0x0000000000403f0e in imp_casign_unprivileged (imp=0x7ffde82a5af0, 
    kv=0xc317e0) at casign.c:129
#7  0x00000000004029d1 in imp_child (ps=ps@entry=0x0, 
    arg=arg@entry=0x7ffde82a5af0) at imp.c:233
#8  0x0000000000402763 in main (argc=2, argv=0x7ffde82a5c18) at imp.c:98
```

garlick:
Ah, the problem is that libtoml keeps doesn't protect itself against too large of an input file.  When reading /dev/zero, eventually the `int` offset pointers wrap around and cause segfaults.  I think we'll just want to impose a limit on cert input size like you did with the kz input.

garlick:
oops, that new function `freads_limited()` would crash if handed /dev/zero!  I'll fix and add a check for that too.

garlick:
OK, fixed that and squashed.

grondo:
This looks good. Merging.

PR #50, opened 2018-02-09
imp: Allow config file pattern to be overridden for build-tree IMP

grondo:
Posting an early version of this (no tests yet) in case @garlick wants to base casign tests off the result (and give early feedback). Will try to work on a few tests tonight, though I just remembered there is a school function, so perhaps not until tomorrow AM.

Sorry this functionality wasn't here from the start.

### DIFF
--- a/src/imp/Makefile.am
+++ b/src/imp/Makefile.am
@@ -31,15 +31,17 @@ flux_imp_SOURCES = \
 	sudosim.h \
 	version.c \
 	whoami.c \
+	testconfig.h \
 	testconfig.c
 
-testconfig.c: $(top_builddir)/config/config.h
+testconfig.o: testconfig.h
+testconfig.h: $(top_builddir)/config/config.h
 	@(confdir=`cd $(srcdir) && pwd`/imp.conf.d; \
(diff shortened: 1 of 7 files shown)


grondo:
Ok, I've pushed some very minor testing updates using the FLUX_IMP_CONFIG_PATTERN support.
While working on this I found some bugs in uses of the imp log functions, so I went ahead and updated those prototypes with printf format attributes and fixed the resulting fallout.

garlick:
Looks good.

I've rebased #48 on this and the only prob was my test under sudo now fails due to missing config file, which sounds right.  I'll just change it to a test_must_fail for now, and when we need a CA as the basis for other tests, we can add an in-tree config file under src/imp, find a place in the tree for ca keys, etc..

Ready to merge this?

grondo:
>  the only prob was my test under sudo now fails due to missing config file, which sounds right. 

Hm, that doesn't sound right to me, the handling of missing config file shouldn't have changed I don't think (but maybe I didn't understand how your test is failing exactly)

garlick:
Fixing my code per offline discussion, and merging this.  Thanks!

grondo:
Thanks!

Issue #52, opened 2018-02-09
 ca_verify should support multiple CA certs


garlick:
When the CA cert is updated, the previous one should remain valid for a time.

New cert signings should use the new cert, but certs signed by the old CA cert should remain valid until they expire.

So perhaps the CA pub key should be stored in a file named for its uuid, and verification should use the "issuer" metadata to map to the CA pub key that should be used for verification.

PR #53, opened 2018-02-14
improve certificate metadata

### LINKED ISSUE #49: add metadata to certificates
Per meeting discussion, we should add the following metadata to user certificates:
```toml
algorithm = "ed25519"
issuer = "uuid-of-ca-cert"
domain = "hype.llnl.gov"
```

garlick:
This PR expands the metadata included in each Flux certificate.

The CA cert is now self-signed through the same process that user certs are signed, so it now includes the same metadata:

* **uuid** -- a unique id for the cert
* **userid** -- real userid of the user that signed the cert
* **ctime** -- certificate create time
* **xtime** -- certificate expiration time
* **domain** -- (new) domain name string from the CA configuration, e.g. `HYPE.LLNL.GOV`
* **issuer** -- (new) the uuid of the cert that signed this cert
* **max-sign-ttl** -- max TTL for signatures by this cert (ignored in the CA cert for now)
* **algorithm** -- (new) always `ed25519` right now

The uuid/issuer relationship prepares for CA key management (#52), where the `issuer` in a user cert can be used to find one of several CA certs with matching `uuid` to validate it.

In this PR, the ctime and xtime are ignored in the CA cert.  CA cert expiration should be worked on in the context of #52 IMHO.

Fixes #49

### DIFF
--- a/src/imp/casign.c
+++ b/src/imp/casign.c
@@ -104,7 +104,7 @@ static void sign_cert (cf_t *conf, struct sigcert *cert)
         imp_die (1, "casign: ca_create: %s", error);
     if (ca_load (ca, true, error) < 0)
         imp_die (1, "casign: ca_load: %s", error);
-    if (ca_sign (ca, cert, ttl, userid, error) < 0)
+    if (ca_sign (ca, cert, 0, ttl, userid, error) < 0)
         imp_die (1, "casign: ca_sign: %s", error);
     if (sigcert_fwrite_public (cert, stdout) < 0)
         imp_die (1, "casign: write stdout: %s", strerror (errno));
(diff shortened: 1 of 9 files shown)


grondo:
Cool, this looks useful! Would be nice to get @dun's review and ACK on these changes as well.

dun:
It might be worthwhile to add a **not-valid-before-time** field to the certificate metadata. This would be useful when transitioning to a new key. It would allow the new key to be distributed before it was valid, and before the existing/old key had expired, but without needing a time interval in which both the old & new keys were valid.

Also, the metadata should have a capability flag that specifies whether the key being certified is a CA. This would allow for a multilevel certificate structure where the CA functionality is split into a hierarchy.

garlick:
Good suggestions - thanks!  I will rework this PR to include both of those.

garlick:
I added the two new cert metadata items, though without full API support at this time.  The **not-valid-before-time** is always set equal to **ctime** (though independently checked), and **ca-capability** is set when the cert is generated by `ca_keygen()`, and checked in `ca_load()`.

There are a couple of commits in this PR that probably should come out and go in a separate one.

I'll do that, add a couple of tests and squash...

garlick:
OK, squashed incremental development and improved test coverage.  I think this is maybe ready to be merged.

Issue #45, opened 2018-02-02
Design input for the IMP exec subcommand


grondo:
Format and contents of the input for the IMP `exec` subcommand still remains to be fully designed.

Because of the IMP's privsep interface, we can define the input in terms of required and optional keys and values, since the format that the "privileged parent" (pIMP) takes is already in the well-defined `struct kv` format. This allows the "unprivilged child of IMP" (chIMP) to be more flexible in the exact format of its input. In fact it may make sense to have multiple exec input "readers" that can be loaded by the chIMP, as long as the data can be formatted into the `struct kv` that the parent expects. (For instance, initially a "testing" format could be used. Future implementations may be able to load a "reader" from their enclosing instance).

grondo:
From RFC 15, required input to the IMP includes

 * Input from instance owner:
   - **R<sub>local<sub>** - Table identifying the locally assigned resources 
   - **options** - Table of key,value pairs identifying optional behavior of the IMP requested by the instance
 * Signed input from guest user (**J**):
    - **security envelope** - security implementation dependent security/signature/credential information (should be transparent to the IMP, decoded as part of verification)
    - **jobspec** - jobspec per RFC14
    - **options** - table of key,value pairs identifying optional behavior of the IMP requested by the user
    - **receipient** - Intended receipient of job request (should equal instance owner, uid running the IMP)
    - **guest userid** - may be part of security envelope
    - **uuid** - unique identifier for this job request
    - **job shell path** - may be part of jobspec(?)
    
Strategy for **J** at this time may be JWT or similar dotted base64 as implemented in the `signer` interface from the flux-security project, and this data will be passed to the privileged parent under a single key, likely just `J = x.y.z`.

This leaves just the design of the "instance owner input to the IMP".

For the purposes of starting simple, I propose that **R<sub>local<sub>** initially take the form of a table of resource types (e.g. core/cpu, memory, gpu, etc), with each table entry comprising a list of logical ids granted to the job's resource set. This will hopefully keep initial implementation very simple.

E.g., in JSON a job using cores and memory might have a **R<sub>local<sub>** that looks something like:
```json
{ 
  "cpu": [  0,1 ],
  "memory":  [ "2G" ],
}
```

[...]

PR #58, opened 2018-02-27
implement flux security context

garlick:
As discussed in #45, this PR implements a simple security context for flux-security.
It also installs libflux-security.so, and moves the libsodium public key stuff over to libca
(the idea is that it will be one of multiple security mechanisms supported, so isolating it
kind of makes sense).
```c
typedef struct flux_security flux_security_t;

flux_security_t *flux_security_create (int flags);
void flux_security_destroy (flux_security_t *ctx);

const char *flux_security_last_error (flux_security_t *ctx);
int flux_security_last_errnum (flux_security_t *ctx);

// NULL glob pattern falls back to compiled-in install path
int flux_security_configure (flux_security_t *ctx, const char *pattern);

int flux_security_aux_set (flux_security_t *ctx, const char *name,
                           void *data, void (*free)(void *));

void *flux_security_aux_get (flux_security_t *ctx, const char *name);
```
Before going much further, wanted to get some feedback on the API so far.


grondo:
Ok, this seems like a good start to me.

garlick:
Hmm, I guess we need accessor fucntions for the parsed configuration.  It seems like those functions should be internal to flux-security so that it's not necessary to expose `cf_t` in the public API.

I guess I'll create a "private" header and not give those functions a `flux_` prefix... (Unless others have better ideas...)

garlick:
Just forced a push with some changes.  I think this is ready for a review (and BTW, I do not have strong opinions about anything in here really, so if anything feels a bit wrong let's discuss)

A shared library is now built and installed containing the `flux_` prefixed functions discussed above.
In addition, some `security_` prefixed functions are exported just to the flux-security project, defined ni context_private.h:
```c

/* Capture errno in ctx->errno, and an error message in ctx->error.
 * If 'fmt' is non-NULL, build message; otherwise use strerror (errno).
 */
void security_error (flux_security_t *ctx, const char *fmt, ...);

/* Retrieve config object by 'key'.
 * Returns the object (do not free), or NULL on error.
 */
const cf_t *security_get_config (flux_security_t *ctx, const char *key);
```

This will allow, for example, a sign/verify implementation to access the parsed configuration, and throw errors that work with the context handle.

Test coverage is better and as a result and a few minor bugs were fixed.

PR #59, opened 2018-03-06
implement generic sign/verify and "none" mechanism

garlick:
This is an attempt at a simple generic sign/verify interface for flux-security.
It is very much like the `libca/signer.h` interface, except it makes use of the `flux_security_t` context for error handling and persistent state.  The main two functions are:

```c
const char *flux_sign_wrap (flux_security_t *ctx,
                            const void *payload, int payloadsz,
                            int flags);

int flux_sign_unwrap (flux_security_t *ctx, const char *input,
                      const void **payload, int *payloadsz,
                      int64_t *userid, int flags);
```

Internally the mechanism is abstracted with callback functions.  There is currently only one mechanism, "none", which appends "none" as the signature and stores the userid in the security header.  Verification succeeds as long as the signature is "none" and the real userid of the unwrapper matches the userid in the security header.

Like the "signer" class, this implementation tries to minimize dynamic memory allocation and perform well when many payloads are "wrapped" and "unwrapped" in succession using the same context.

If unwrap is called with the FLUX_SIGN_NOVERIFY flag, it avoids rechecking the signature and simply hands over the payload and userid even if the signature is not valid (e.g. if it's known to have been verified already without opportunity for tampering).

I don't feel like this is quite done yet, but I wanted to get it out there in case there is general feedback.  I'll need to add test coverage, and that may require some new interfaces such as the ability to register a new mechanism dynamically, which I avoided writing the code for thus far.


grondo:
Looks fine so far. I like the simplicity of the wrap/unwrap interface.

Before landing this it might be useful to walk through what the "plugin" for the munge and CA mechanisms will look like, just to ensure the internal interface works for those cases.

garlick:
I have some changes to the mech-specific interface plus various cleanups queued up for this morning.  I'll push those (to be squashed later) and then post a sketch of what the mechs would look like for munge and fluxca.  I think I've got that worked out in my head but it would be useful to be a bit more methodical.

garlick:
Under the proposed abstraction, a munge mech would implement the following operations:

**init**

* verify `[sign.munge]` config (long-term TTL, etc)
* create munge context
* stash munge context + config (a local context) in `flux_security_t` for later use

**prep**

Not implemented

**sign**

* retrieve local context
* compute SHA1 over HEADER.PAYLOAD input
* munge_encode SHA1 payload
* cache munge cred in local context (for memory management) and return it as a `const char *`

**verify**

* retrieve local context
* compute SHA1 over HEADER.PAYLOAD portion of input
* munge_decode SIGNATURE part of input, ignoring EMUNGE_CRED_REPLAYED or EMUNGE_CRED_EXPIRED errors
* verify payload SHA1 matches computed
* verify munge userid matches header userid
* munge_opt_get MUNGE_OPT_ENCODE_TIME from munge context to get last cred encode time
* verify encode time > wallclock (within error bars)
* verify encode time + configured TTL >  walllclock (within error bars)

grondo:
> * cache munge cred in local context (for memory management) and return it as a const char *
 
I'm assuming the "memory management" is to avoid a copy of the result? However, does every signature created persist in the security handle until the security handle is destroyed?

garlick:
Since`flux_sign()` immediately copies the result returned by mech->sign, it really doesn't matter whether `flux_sign()` frees the result, or it gets overwritten by the next call to the mech->sign callback, which is what I had implemented.  For the other methods, I was starting with a const char *, so to do it the other way I would have to strdup it.  Either way is fine with me, if you have a preference.

garlick:
A fluxca mech would implement the following operations

**init**

* verify `[sign.fluxca]` config (TTL etc)
* store config (local context) in `flux_security_t` for later use

**prep**

* retrieve local context
* `sigcert_load()` users public/private certs and add to local context
* add fluxca.cert to header
* add fluxca.ctime = wallclock to header
* add fluxca.xtime = wallclock + configured TTL to header

**sign**

* retrieve local context
* `sigcert_sign_detached()` HEADER.PAYLOAD portion of input
* return base64 signature

**verify**

* retrieve local context
* verify fluxca.ctime <= wallclock
* verify fluxca.xtime > wallclock
* verify fluxca.xtime - fluxca.ctime <= configured max TTL
* `sigcert_verify_detached()` HEADER.PAYLOAD against SIGNATURE portion of input

grondo:
> Since flux_sign() immediately copies the result returned by mech->sign, it really doesn't matter whether flux_sign() frees the result, or it gets overwritten by the next call to the mech->sign callback, which is what I had implemented. 

Sorry, I hadn't looked in any detail at the mechanism yet. Let me at least do that before commenting further.

grondo:
I might be making a problem where there isn't one, but it appears that signing is not MT-safe unless you create a separate flux_security_t handle per thread? Additionally, signing multiple blobs without copying the results is not allowed? I guess all I'm proposing is that it might result in fewer future difficult-to-find errors if `flux_sign_wrap` returned memory that has to be freed by the caller?

I guess it makes sense that there is only one allowed signer mechanism (based on configuration) on the signer side, on the verification side I could imagine there might be trouble if we stick with the "there can be only one" `struct sign` per `flux_security_t` object. Eventually we may want verification to accept one of a set of supported mechanisms. (?)

garlick:
I thought it would be OK to require a separate `flux_security_t` context per thread, same as a munge context, or `flux_t`.  Context interfaces like `flux_security_last_error()` don't really work unless you have one per thread, so we're pretty off track if that is something we need...

My thought on  the `flux_sign_wrap()` return result and the similar payload handling in `flux_sign_unwrap()` was that this way simplifies the "error unwind" code for callers, is less prone to leaks, and allows the result buffers to be reused, thus reducing calls to malloc and free.  I might be making more out of that than is justified though.   Still convinced it should go the other way?   LMK - I don't mind reworking it.

garlick:
On the allowed mechanisms - you are probably right.  I'll tweak the config so the verifier end can allow multiple mechanisms.

grondo:
>  thought it would be OK to require a separate flux_security_t context per thread, same as a munge context, or flux_t. Context interfaces like flux_security_last_error() don't really work unless you have one per thread, so we're pretty off track if that is something we need...

Oh yeah, you are correct here, not sure what I was thinking...

> My thought on the flux_sign_wrap() return result and the similar payload handling in flux_sign_unwrap() was that this way simplifies the "error unwind" code for callers, is less prone to leaks, and allows the result buffers to be reused, thus reducing calls to malloc and free. I might be making more out of that than is justified though. Still convinced it should go the other way? LMK - I don't mind reworking it.

Well my initial worry when looking at the code would only be a concern for a case where a caller was signing 2 or more separate objects, it might be tempting to call `flux_sign_wrap()` multiple times before copying the data into a message or other container. The resulting errors might be very difficult to track down.

If this isn't an expected use case I'm willing to try it this way for now.

grondo:
> Oh yeah, you are correct here, not sure what I was thinking...

Ah, I remember now. I was (erroneously) thinking of a single `flux_security_t` accessed by all threads so that configuration would not have to be opened, read, and parsed by every thread in a process. Perhaps the solution to that would be as simple as offering a copy operation for `flux_security_t`.

garlick:
I'll open a bug on `flux_security_copy()` for that use case - good idea!

I pictured the main use cases for sign_wrap/unwrap are 
* iteratively creating job requests (signed result likely incorporated into a message right away)
* iteratively receiving job requests and storing them somewhere (payload parsed into object and discarded?)
* imp probably receiving one job request per invocation?

It's hard to anticipate all the ways it will be used.  if you're OK with this way for now, I hereby declare there will be no whining on my part if use cases emerge that prove your point later on.

garlick:
Just forced a push that squashes earlier incremental development, expands commit messages, adds config support for allowing multiple unwrap mechanisms, and adds inline docs to sign.h.  Unit test updated to cover additional config errors that are now possible.

If coverage is still OK and the changes look good, I should squash one more time.

grondo:
The commit message in 2efef3b is really instructive, thanks! IMO you should put that in either `signer.h` or `signer.c` so that it is more readily available to developers perusing the code. (at least the description of the "wrapped" signature format would be really useful in `sign.h`)

garlick:
Glad it's helpful!  I will do that.

grondo:
My only other comment is that I think we should make a pass through the configuration keys in the eyes of an admin. E.g. I'm not sure the wrap/unwrap terminology (without being familiar with the code) conveys the fact that the configuration is affecting methods for generating and verifying signatures. It might be more clear as something like:
```
[sign]
  allowed-types = [ munge, ca ]
  default-type = "munge"
  max-ttl = 12356
[sign.munge]
  socket = "/path/to/munge.sock"
  zip = "bzlib"
```
(did I get the sub-table right for toml here?)

However, I don't want to change the config just for the sake of changing it now, since it will likely have to change again as we get further down the road, but in general I just wanted to be aware of this comment.

garlick:
Ah, that's a very good comment.  I really was not thinking about it from that perspective.

I think the "allowed-types" and "default-type" keys are much better than "unwrap-mechanisms" and "wrap-mechanism".   I'll go ahead and switch to that now, and we can always change it again later.

> did I get the sub-table right for toml here?
Yep!

See also: [Table](https://github.com/toml-lang/toml#table)

garlick:
I'll squash the whole thing in a bit - seems like it's getting close.

grondo:
Thanks! this seems to be in pretty good shape.

For the use case of the "none" mechanism, I'm kind of wondering how a single-user instance could switch to this signing mechanism as the default after initialization. For example, I've run a batch job (new instance) within the system instance using munge as dictated by system configuration. Within the batch script (i.e. initial program) of this new, single-user instance, I issue `flux run` which by default also looks at system configuration and will default to "munge". In fact, it doesn't seem like `flux run` will know if the `none` mechanism could be used until it contacts the instance and queries for the instance user. I guess this isn't important at this time, use of the "munge" signature will also work fine for single-user instances (maybe "none" is reserved for testing or launch of single user instance under another resource manager or out of a builddir)

This question is outside the scope of this PR (which seems ready to merge IMO), but I guess what I'm getting at is we don't have a mechanism yet to configure the wrap side of the interface yet, and I wonder if that will be needed (perhaps I should open a new issue on this if I can actually phrase my question to make any sense!)

garlick:
That's probably a good issue to open.   I was thinking a user running their own instance might point flux at a config other than the system one, but that might lead to submitting with a mechanism the IMP doesn't accept.

Hmm definitely needs more thought/discussion.

This is ready as far as I'm concerned.  Thanks for all the helpful review comments.

Issue #66, opened 2018-03-12
t1002-sign-munge fails on Ubuntu 16.04 LTS


garlick:
If munged is not new enough to work with `03-munged.sh`, test should SKIP rather than FAIL.

In addition, if older munge is configured and running, tests can use it without being held up by the missing options needed to start/stop locally.

Finally, the expected failures should probably expect more specific errors, so that they do not appear to "pass" when some other failure masks the result.
```
$ ../t1002-sign-munge.t --verbose
munge-0.5.11 (2013-08-27)
expecting success: 
	munged_start_daemon

munged: Error: Invalid option "--pid-file=/home/garlick/proj/flux-security/t/trash-directory.t1002-sign-munge/munged.pid.4889"
not ok 1 - start munged
#	
#		munged_start_daemon
#	

expecting success: 
	munge --socket ${MUNGE_SOCKET} </dev/null \
		| unmunge --socket ${MUNGE_SOCKET} >/dev/null

munge: Error: Missing argument for option "--socket"
unmunge: Error: Missing argument for option "--socket"
not ok 2 - munge works
#	
#		munge --socket ${MUNGE_SOCKET} </dev/null \
#			| unmunge --socket ${MUNGE_SOCKET} >/dev/null
#	

expecting success: 
	cat >sign.toml <<-EOT
	[sign]
	max-ttl = 60
	default-type = "munge"
	allowed-types = [ "munge" ]
	[sign.munge]
	socket-path = "${MUNGE_SOCKET}"
	EOT

ok 3 - create sign.toml

expecting success: 
	cat /dev/null >zsign.in &&
	${sign} <zsign.in >zsign.out &&
	${verify} <zsign.out >zverify.out &&
	test_cmp zsign.in zverify.out

sign: flux_sign_wrap: sign-munge-sign: MUNGE socket name is undefined
not ok 4 - sign/verify zero length payload
#	
#		cat /dev/null >zsign.in &&
#		${sign} <zsign.in >zsign.out &&
#		${verify} <zsign.out >zverify.out &&
#		test_cmp zsign.in zverify.out
#	

expecting success: 
	echo Hello >sign.in &&
	${sign} <sign.in >sign.out &&
	${verify} <sign.out >verify.out &&
	test_cmp sign.in verify.out

sign: flux_sign_wrap: sign-munge-sign: MUNGE socket name is undefined
not ok 5 - sign/verify a short message
#	
#		echo Hello >sign.in &&
[...]

PR #67, opened 2018-03-12
sign_munge: fix memory leak, improve test resiliency

garlick:
This PR addresses the`t1002-sign-munge.t` limitations discussed in #66, fixes a memory leak in the munge mechanism implementation, and fixes a typo in `sign.h` header docs.


t/t1002-sign-munge.t
```diff
@@ -96,12 +106,17 @@ test_expect_success 'message with altered munge cred fails verify' '
...
 	[sign]
 	max-ttl = -100
 	default-type = "munge"
 	allowed-types = [ "munge" ]
+	EOT
+'
+
+test_expect_success SIDEMUNGE 'adding munge.socket-path to sign.toml' '
+	cat >>sign.toml <<-EOT
 	[sign.munge]
```
grondo: Instead of skipping configuration test here, could we add the system munge socket path explicitly?
Could this allow one of the redundant `'adding munge.socket-path'` "tests" to go away?

grondo: Meant to add that system munge socket-path is available via `munged --help | grep socket= | sed 's/.*\[\(.*\)\]$/\1/'
`

garlick: That was my first thought too, but I'm not sure how to reliably determine the path?

grondo: I think if `munge | unmunge` works with setting `--socket`, then it can be assumed that the MUNGE daemon is using the default socket-path, which is printed in `--help` output (see above)

garlick: Got it, thanks!

grondo:
Ah, yes the test output is much clearer too! Looks good to me, ready to merge?

garlick:
yep!

PR #68, opened 2018-03-13
sign_curve: add curve signing mechanism

garlick:
This is a first stab at reworking the libca certificate/ca code into a `sign_curve` mechanism.

In addition to verifying that enclosed certiifcates were signed by the CA and thus authenitcate the claimed userid, if `require-ca = false` is configured, an identical certificate present in  the user's `$HOME/.flux/curve/sign.pub` is sufficient to validate the user.

The old signer class will be scrapped and its unit tests repurposed to make a utility like `xsign_munge.c` which can create improperly signed messages of various kinds, driven from sharness.

For testing (to avoid having to access a test user's home directory), the signing key can be fored to a particular path with the `cert-path` configuration.  The sharness tests use this.


garlick:
Whew, finally got the test coverage of this diff up to a reasonable level.

I'd say this is probably ready for a review.

t/Makefile.am
```diff
@@ -33,7 +33,8 @@ check_PROGRAMS = \
 	src/cf \
 	src/sign \
 	src/verify \
-	src/xsign_munge
+	src/xsign_munge \
+	src/xsign_curve
```
grondo: commit message references `xsign_munge`, but utility is `xsign_curve`, typo?

t/src/verify.c
```diff
@@ -47,7 +47,7 @@ static int read_all (void *buf, int bufsz)
 int main (int argc, char **argv)
 {
     flux_security_t *ctx;
-    char buf[1024];
+    char buf[8192];
     int buflen;
```
grondo: It is minor, but typo in the commit message `s/fix/fixed/`

grondo:
Hmm, review lost one of my comments. Commit message in 00778e6: `s/xsign_munge/xsign_curve/`

garlick:
Fixed those and forced a push.  Thanks!

garlick:
I'll go ahead and make the small changes to the signing mech interface proposed in #70 (from review in #63) in this PR, standby.

garlick:
Re-pushed with the fixes to mech interfaces discussed #70.

I also made the mech->sign() callback return a heap allocated string as suggested by @grondo earlier.   (I guess that one took a while to sink in!)

Also dropped the "attached signature" functions in sigcert, which are no longer used now that the signer class is dropped.

garlick:
Rebased.  I think this was in a mergeable state.

grondo:
Yeah, I was wondering why we held off on merging this. Sorry it must have just fallen off the radar.

PR #69, opened 2018-03-14
enhance private context interface and other small fixes

grondo:
This is a list of small fixes split off of ongoing work. I wanted to submit these before they got too out of date.

The main functionality changes are minor and to only the private security context interface in `context_private.h`. The allow an existing `cf_t` config to be set directly in `flux_security_t` with a new `security_set_config()` call, and add a way to return the entire `cf_t` stored in the `flux_security_t` context by passing a `NULL` as the key in `security_get_context`.

These are needed to allow the IMP, which parses its own config, to share all the config tables it has with the security handle.

The minor fixes were found along the way and include fix for off-by-one error and leak found in sign.c, and a small typo in cf_copy() test.

### DIFF
--- a/src/lib/context.c
+++ b/src/lib/context.c
@@ -214,6 +214,24 @@ const cf_t *security_get_config (flux_security_t *ctx, const char *key)
 
 }
 
+int security_set_config (flux_security_t *ctx, const cf_t *cf)
+{
+    cf_t *new;
+    if (!ctx || !cf) {
+        errno = EINVAL;
+        security_error (ctx, NULL);
+        return (-1);
+    }
+    if (!(new = cf_copy (cf))) {
+        errno = ENOMEM;
+        security_error (ctx, "Failed to copy config object");
+        return (-1);
+    }
+    cf_destroy (ctx->config);
+    ctx->config = new;
+    return (0);
+}
+
 /*
  * vi:tabstop=4 shiftwidth=4 expandtab
  */
--- a/src/lib/context.c
+++ b/src/lib/context.c
@@ -194,7 +194,7 @@ const cf_t *security_get_config (flux_security_t *ctx, const char *key)
 {
     const cf_t *cf;
 
-    if (!ctx || !key) {
+    if (!ctx) {
         errno = EINVAL;
         security_error (ctx, NULL);
         return NULL;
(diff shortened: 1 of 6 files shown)


grondo:
BTW, I also have a patch that simplifies `test/sign.c` using `security_set_config` to directly load test `cf_t` from memory instead of dumping configuration to files and loading with `flux_security_configure`. However, I thought better of submitting that because it replaces use of a public function (`flux_security_configure`) with a private function (`security_set_config`) which wouldn't be used by callers. Also, the change seemed wanton and unnecessary since the tests are working fine as is.

grondo:
Not much tested code here, so the unchecked error handling code in the new context.c function is causing low coverage here.

garlick:
Thanks for finding the leak and the off-by-one bug!

Looks good - ready?

grondo:
I added one more test  here, but could pop it off if you'd like.

I noticed an untested code path in codecov results for rejection of messages encoded with a type that doesn't appear in `allowed-types`. This last commit just checks that the decode/unwrap fails in the expected way (piggybacking on the t1002-sign-munge tests).

garlick:
Hmm, that should have been covered by this unit test from src/lib/test/sign.c:
```c

    const char *badconf_unknown_default_type = \
    "[sign]\n" \
    "max-ttl = 32\n" \
    "default-type = \"foo\"\n" \
    "allowed-types = [ \"none\" ]\n";

    if (!(ctx = context_init (badconf_unknown_default_type)))
        BAIL_OUT ("failed to set up test config");
    errno = 0;
    ok (flux_sign_wrap (ctx, "foo", 3, 0) == NULL && errno == EINVAL,
        "flux_sign_wrap with unknown default-type config fails with EINVAL");
    diag ("%s", flux_security_last_error (ctx));
    flux_security_destroy (ctx);
```
was it not showing up green in the coverage tests?

grondo:
It is the unwrap side that wasn't showing as covered. I.e. the case where a validly encoded signature is rejected because the encode type isn't in allowed-types. (That is, the `unwrap` side, not `wrap` side)

I'm sorry I didn't go through and check existing tests to see if the coverage report was wrong or there was a test that wasn't working exactly right (should have done that!)

In any event, it probably does make more sense to test this case in the unit test, but it seemed a lot easier to piggyback on the existing `sign` `verify` executables instead.

grondo:
Here's the untested code fragment, so you don't have to go searching:
https://codecov.io/gh/flux-framework/flux-security/src/master/src/lib/sign.c#L445

garlick:
Oh, right!  That makes sense then.  Your test is fine where it is as far as I'm concerned.  Thank you!  OK to merge?

grondo:
sure thanks!

Issue #62, opened 2018-03-07
How to select different signer mechanism for flux_sign_wrap?


grondo:
Flux utilities making use of `flux_sign_wrap` will need to read default system `[sign]` config table in order to properly create signed data in a form accepted by the system installed IMP. This is straightforward when interacting with a system instance, since all system-wide config should apply.

However, if a child instance wants to allow tools/utilities to use an alternate signer mechanism for `flux_sign_wrap` there is currently no way to do this since only one `sign.default-type` is stored in configuration. For instance, a single user instance may be able to use the "none" signature mechanism instead of the more secure mechanism of the system instance `sign.default-type`.

Users of `flux_security_t` and `flux_sign_wrap` should somehow be able to select an alternate `sign.default-type`, either via exposing the `flux_security_t` config values in some way, or as another parameter to `flux_sign_wrap`. Callers of this interface should get a warning if the selected mechanism isn't supported by system configuration.

Another question would be where and how a tool like `flux run` would decide that it could fall back to "none" signer mechanism. Perhaps a requirement for some sort of "discovery" for these tools should be written into global flux requirements/plans.

Of course, this isn't a critical issue now since it doesn't hurt to use the more secure system-configured `sign.default-type`, but thinking about how the interface would look might be illustrative.

garlick:
Adding a `const char *mechanism` argument to `flux_sign_wrap()` seems entirely reasonable.  Then a command could override the default mechansim if it wants.

I'm starting to wonder if maybe "unwrap" should do its best for any of the known mechanisms, provide a way to tell which mechanism was used, and then let the caller apply "allowed-types" after the fact, since in addition to site policy, the context is important too (who is calling unwrap and do we have other knowledge such as a peercred that makes e.g. mechanism=none acceptable).

grondo:
That seems perfectly reasonable to me. An IMP being run where ruid == target uid could also accept mechanism==none.

grondo:
> I'm starting to wonder if maybe "unwrap" should do its best for any of the known mechanisms, provide a way to tell which mechanism was used, and then let the caller apply "allowed-types" after the fact, since in addition to site policy, the context is important too (who is calling unwrap and do we have other knowledge such as a peercred that makes e.g. mechanism=none acceptable).
￼
On the other hand, flux-security APIs should probably be designed to "fail securely" and it might be too easy for the casual programmer to "forget" to apply the allowed-types after the fact and create a security hole.

[...]

PR #75, opened 2018-06-05
sign: add mechanism arguments

garlick:
As discussed in #62, this PR adds a mechanism argument to `flux_sign_wrap()`:
```c
// mech_type=NULL means use configured 'default-type'
const char *flux_sign_wrap (flux_security_t *ctx,
                            const void *payload, int payloadsz,
                            const char *mech_type,
                            int flags);
```
and a new function `flux_sign_unwrap_anymech()`:
```c
int flux_sign_unwrap_anymech (flux_security_t *ctx, const char *input,
                              const void **payload, int *payloadsz,
                              const char **mech_type,
                              int64_t *userid, int flags);
```

I need to add some tests.


garlick:
Cleaned up a bit (squashed) and updated unit tests to cover the diff.

src/lib/sign.c
```diff
@@ -281,20 +277,29 @@ const char *flux_sign_wrap (flux_security_t *ctx,
     }
     if (!(sign = sign_init (ctx)))
         return NULL;
+    if (!mech_type)
+        mech_type = cf_string (cf_get_in (sign->config, "default-type"));
+    if (!validate_mech (ctx, mech_type, &mech))
```
grondo: Question:
If `!mech_type` then is `validate_mech()` redundant here? Could maybe be a crash if `default-type` not set in `sign->config` and `NULL` mech_type` is passed to `validate_mech()`?

grondo: following on that same thought `validate_mech()` should probably just return false or assert() if name == NULL.

grondo:
LGTM! Just one minor comment inline...

src/lib/sign.c
garlick: Yep, `validate_mech()`should be changed to `lookup_mech()` since the mech was already validated in `sign_create()`.   Fix coming.

If `default-type` is not set, the `cf_check()` call in `sign_create()` will fail, since `.required = true` for that key in `sign_opts[]`.  So it should be safe to access it without checking later on.

garlick:
> Yep, validate_mech()should be changed to lookup_mech() since the mech was already validated in sign_create(). Fix coming.

Ooops, actually the validate is needed.  Only the configured mechs are "validated", but the user can specify one that's neither `default-type` nor in `allowed-types`.

So I think it's actually correct as is.

grondo:
> Ooops, actually the validate is needed. Only the configured mechs are "validated", but the user can specify one that's neither default-type nor in allowed-types.

I was only talking about the case where the user didn't specify a `mech_type`. You then look up  the `default-type`, then validate it. This is confusing on first read -- is is possible for the `default-type` to not be in `allowed-types`? I was actually just wondering if there was wasted work in the case where `NULL` is passed to `sign_wrap` (which will likely be by far the common case)

> So I think it's actually correct as is.

src/lib/sign.c
grondo: > If default-type is not set, the cf_check() call in sign_create() will fail, since .required = true for that key in sign_opts[]. So it should be safe to access it without checking later on.

Ah, ok. Makes sense. Also, as long as `validate_mech()` is never called with data from the user it should be ok to skip a check for NULL argument.

Also looks like `validate_mech` is always called on the `default-type` in `sign_create` -- I guess it doesn't hurt to do it again when a user calls `flux_sign_wrap()` using `mech_type == NULL`, but since this is going to be the common case I thought why not drop it (sorry if I wasn't clear in my initial question)?

grondo:
Yes, thanks. Looks like now the default-type isn't doubly verified, though I know it was a minor concern!

garlick:
Travis was not cooperating (lots of failures fetching packages), but it finally passed.

I would vote we merge this.

grondo:
Ok, seems good!

PR #76, opened 2018-06-11
etc: install config file for signing

garlick:
This PR moves the default config file location to `$sysconfdir/flux/security/conf.d`, and installs a default config file for signing.

The config file location change keeps flux-core and flux-security configs separate so that security components like the IMP don't parse irrelevant (and possibly voluminous or conflicting) TOML, and vice versa.

The default config file allows the installed flux-security package to be functional without manual configuration.  At the moment this allows code in flux-core to configure flux-security with a NULL config pattern and get something that works, which is convenient while we evaluate how to share configuration and security contexts within a flux instance.  When flux-security is packaged as an RPM or deb, we would tag this config file as a "noreplace" so that any installed config takes precedence.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,4 +1,4 @@
-SUBDIRS =         src t
+SUBDIRS =         src etc t
 ACLOCAL_AMFLAGS = -I config
 EXTRA_DIST = \
 	config/tap-driver.sh \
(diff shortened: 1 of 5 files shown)


grondo:
Seems reasonable to me!

garlick:
Thanks!

PR #77, opened 2018-06-12
build: add pkg.m4 to support side-install autotools

garlick:
Problem: when building flux-core in travis, the dep-builder
script pulls in flux-security and autogens it, then configure
fails.

This is the same problem as discussed in
flux-framework/flux-sched#242

Pull in pkg.m4.

(Thanks to @grondo for the offline guidance.)


grondo:
Thanks!

Issue #84, opened 2018-11-01
base64: switch to libsodium implementation


garlick:
As discussed in flux-framework/flux-core#1775 and implemented in flux-framework/flux-core#1786, flux is switching to the base64 implementation in libsodium.

flux-security should make this change also.

Issue #82, opened 2018-10-05
remove typedefs that are RFC7 non-compliant


chu11:
Was looking around all of flux to see what code may be non-compliant given recent https://github.com/flux-framework/rfc/pull/139 change.  It appears there's a nice chunk in flux-security.

```
src/libca/ca.c:typedef char uuidstr_t[37];     // see uuid_unparse(3)
src/libca/ca.h:typedef char ca_error_t[200];
src/libca/sigcert.c:typedef uint8_t public_t[BASE64_DECODE_MAXLEN(crypto_sign_PUBLICKEYBYTES)];
src/libca/sigcert.c:typedef uint8_t secret_t[BASE64_DECODE_MAXLEN(crypto_sign_SECRETKEYBYTES)];
src/libca/sigcert.c:typedef uint8_t sign_t[BASE64_DECODE_MAXLEN(crypto_sign_BYTES)];
src/libca/sigcert.c:typedef char public_base64_t[BASE64_ENCODE_LENGTH(crypto_sign_PUBLICKEYBYTES)];
src/libca/sigcert.c:typedef char secret_base64_t[BASE64_ENCODE_LENGTH(crypto_sign_SECRETKEYBYTES)];
src/libca/sigcert.c:typedef char sign_base64_t[BASE64_ENCODE_LENGTH(crypto_sign_BYTES)];
```

PR #85, opened 2018-11-02
switch to libsodium base64 implementation

garlick:
This PR switches flux-security over to the libsodium base64 implementation as requested in #84.  That change eliminated most of the RFC 7 violations called out in #82 except one, which is added on a separate commit.


grondo:
Looks like travis-ci build also needs a libsodium upgrade. I can add this onto your PR if you are away from a computer now.

garlick:
Just pushed that - thanks though!

src/libca/sigcert.c
```diff
@@ -52,34 +52,26 @@
...
-
+#define PUBLICKEY_BASE64_SIZE \
+    (sodium_base64_ENCODED_LEN (crypto_sign_PUBLICKEYBYTES, \
+                                sodium_base64_VARIANT_ORIGINAL))
+#define SECRETKEY_BASE64_SIZE \
+    (sodium_base64_ENCODED_LEN (crypto_sign_SECRETKEYBYTES, \
+                                sodium_base64_VARIANT_ORIGINAL))
+#define SIGN_BASE64_SIZE \
+    (sodium_base64_ENCODED_LEN (crypto_sign_BYTES, \
+                                sodium_base64_VARIANT_ORIGINAL))
```
grondo: I'm probably being overly cautious here, but these function calls masquerading as #define'd constants are troubling me.
Would it be better to add `()` to these macros so any user knows they are function-like macros and not constants?

Since these macros are internal to `sigcert.c` it actually is not too important.

grondo:
This looks good to me!

I had one question about the internal #defines in sigcert.c, I commented on the commit but not sure where that went. After I commented I realized those macros are internal to `sigcert.c` so I'm not sure my comment is too important. (It just looked strange to me to have those macros that look like constants, but are actually calling functions. I would slightly prefer appending `()` so the fact they are function-like macros is evident)

src/libca/sigcert.c
garlick: `sodium_base64_ENCODED_LEN()` is a macro (as opposed to `sodium_base64_encoded_len()` which is not), so these are constant.  They need to be as I'm using them to define static buffers.

garlick:
(Answer to your question was inline)

Hitting the road - will check in again this evening.

src/libca/sigcert.c
grondo: Ah, sorry that I missed that!

> . They need to be as I'm using them to define static buffers.

Actually, C99 supports variable-length arrays so I believe your buffer definitions would work with a function providing the size for intialization.

PR #86, opened 2018-11-09
add flux-security version accessors

garlick:
This adds functions and macros for accessing the flux-security version, as was done for flux-core in flux-framework/flux-core#1817


grondo:
Nice!

PR #88, opened 2018-11-10
lib: make flux_security_aux_get/set behave like flux-core API

garlick:
Following on flux-framework/flux-core#1797, switch the implementation behind `flux_security_aux_set()` and `flux_security_aux_get()` over to the common one from flux-core (and pull that in with its unit test).


garlick:
Rebased on current master.

garlick:
This was rebased BTW, so it should be ready to go

PR #91, opened 2019-01-03
LLNS re-release under LGPL-3.0

garlick:
Flux-framework has been approved for re-release under LGPL-3.0.

This PR to flux-security is similar to flux-framework/flux-core#1901.

I did leave off the LICENSE and NOTICE symlinks as they seem superfluous since we have COPYING and NOTICE.LLNS.  They could be added later if the powers that be decree them necessary.


grondo:
Thanks for doing all that work!

garlick:
Thanks!

PR #97, opened 2019-12-13
Add option to build with asan and ubsan, and fix discovered errors

grondo:
This PR adds an `--enable-sanitizers` option to `configure` which turns on `-fsanitize=address` and `-fsanitize=undefinied` (along with other requisite options), and enables a build with this option in travis-ci.

Along the way some minor fixes (mostly in tests) were discovered and fixed.

At the top-level Makefile, some default ASAN_OPTIONS and UBSAN_OPTIONS are set, so that `make check` from top builddir will run successfully. (However, it won't work from a subdir, so perhaps another method is required)

Some shenanigans were required to work around "issues" with LSan and sudo/setuid.

I can add more detail later, but wanted to get this up to see how Travis goes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,27 +1,22 @@
 language: c
 sudo: true
+dist: bionic
 
 matrix:
   include:
-    - compiler: gcc
+    - compiler: gcc-8
       env: MAKECMDS="make && make distcheck"
     - compiler: clang
-    - compiler: gcc-5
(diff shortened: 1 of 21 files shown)


grondo:
The coverage drop here has me a little concerned. I can't really figure out what happened. [codecov.io](https://codecov.io/gh/flux-framework/flux-security/pull/97?src=pr&el=h1) is reporting that most of the drop is in libtomlc99, so I want to make sure I didn't inadvertently drop a test there or something.

grondo:
Didn't find anything obvious. Part of this PR updated travis builders to Ubuntu bionic to get new compilers, so perhaps part of the coverage change here is due to that update.

grondo:
Thanks!

PR #100, opened 2020-02-25
testsuite: fix tests that do not work in some environments

grondo:
This PR fixes a couple tests in flux-security that are broken on LC clusters.

 Nothing too major:

 * Update the method for determining system munged socket (newer MUNGE versions have another line that contains "socket" in the `--help` output)
 * Skip setuid tests in the IMP testsuite when running on a filesystem with nosuid (or other scenarios where setuid permissions cannot be set on executables)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,12 @@
+on: pull_request
+jobs:
+  check-pr:
+    name: validate commits
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


grondo:
I've pushed a couple other commits here for expediency:
 * add mergify config
 * add GitHub pr-validator action

I'm perfectly willing to submit a separate PR for the mergify and GitHub Workflow additions, since they are unrelated to the testsuite fixes.

grondo:
Thanks!

PR #104, opened 2020-03-06
imp: add limited exec support

### LINKED ISSUE #96: imp: need limited exec prototype
We need to prototype the imp exec function with the following constraints:
* only the flux user (or a configured list of users) is allowed to run imp exec
* only the installed flux job shell (or a configured list of executable paths) may be run by imp exec

grondo:
This PR adds a minimal working version of the `flux-imp exec` subcommand.
This version takes input as JSON on stdin. The current format requires a single key "J", expected to contain the users signed jobspec.

`flux-imp exec` reads the path to the job shell to execute plus 1 argument from the cmdline. This allows simply inserting `flux-imp exec` in front of the job shell invocation already being done by the job-exec module.

With the exec command comes a new `[exec]` table in the IMP's configuration. In order for `flux-imp exec` to be functional, the table must contain at least the following keys
```toml
[exec]
allowed-users = [ "list of users allowed to run imp exec" ]
allowed-shells = [ "list of shell paths allowed for imp exec" ]
```

Without either of these config keys either the IMP user or the job shell will not be allowed and thus `flux imp exec` will not be functional.

Additional changes of note:
 - The flux-imp build was split into an installed IMP (`inst/flux-imp`) and test IMP `./flux-imp`. The only difference is how the IMP reads its config. The test IMP's config can be overridden at runtime, and it reads IMP and libflux-security config from the same config pattern.

Closes #96

### DIFF
--- a/src/imp/passwd.c
+++ b/src/imp/passwd.c
@@ -1,0 +1,60 @@
+/************************************************************\
+ * Copyright 2020 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include <stdlib.h>
+#include <string.h>
+
+#include "passwd.h"
+
+static struct passwd * passwd_copy (struct passwd *arg)
+{
+    struct passwd *pwd = calloc (1, sizeof (*pwd));
+    if (pwd) {
+        pwd->pw_uid = arg->pw_uid;
+        pwd->pw_gid = arg->pw_gid;
+        if (!(pwd->pw_name = strdup (arg->pw_name))
+            || !(pwd->pw_passwd = strdup (arg->pw_passwd))
+            || !(pwd->pw_gecos = strdup (arg->pw_gecos))
+            || !(pwd->pw_dir = strdup (arg->pw_dir))
+            || !(pwd->pw_shell = strdup (arg->pw_shell))) {
+            passwd_destroy (pwd);
+            return NULL;
+        }
+    }
+    return pwd;
+}
+
+struct passwd * passwd_from_uid (uid_t uid)
+{
+    struct passwd *pwd = NULL;
+    if (!(pwd = getpwuid (uid)))
+        return NULL;
+    return passwd_copy (pwd);
+}
+
+void passwd_destroy (struct passwd *pwd)
+{
+    if (pwd) {
+        free (pwd->pw_name);
+        free (pwd->pw_passwd);
+        free (pwd->pw_gecos);
+        free (pwd->pw_dir);
+        free (pwd->pw_shell);
+        free (pwd);
+    }
+}
+
+/*
+ * vi: ts=4 sw=4 expandtab
+ */
--- a/src/imp/passwd.h
+++ b/src/imp/passwd.h
@@ -1,0 +1,28 @@
+/************************************************************\
+ * Copyright 2020 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+#ifndef HAVE_IMP_PASSWD_H
+#define HAVE_IMP_PASSWD_H 1
+
+#include <pwd.h>
+#include <sys/types.h>
+
+/*
+ *  Return a copy of the passwd entry for UID
+ *  Caller must free with passwd_destroy()
+ */
+struct passwd * passwd_from_uid (uid_t uid);
+
+/*
+ *  Free memory for a copy of passwd entry created by passwd_from_uid
+ */
+void passwd_destroy (struct passwd *pw);
+
+#endif /* !HAVE_IMP_PASSWD_H */
--- a/src/imp/privsep.c
+++ b/src/imp/privsep.c
@@ -164,26 +164,29 @@ privsep_t * privsep_init (privsep_child_f fn, void *arg)
     return (ps);
 }
 
-int privsep_destroy (privsep_t *ps)
+int privsep_wait (privsep_t *ps)
 {
     int status = 0;
-
-    if (ps->wfd > 0)
-        close (ps->wfd);
-    if (ps->rfd > 0)
-        close (ps->rfd);
-
     if (privsep_is_parent (ps)) {
         if (ps->cpid > (pid_t) 0) {
-            int status;
             kill (SIGTERM, ps->cpid);
             if (waitpid (ps->cpid, &status, 0) < 0)
                 status = -1;
+            ps->cpid = (pid_t) -1;
         }
     }
-
-    free (ps);
     return (status == 0 ? 0 : -1);
+}
+
+void privsep_destroy (privsep_t *ps)
+{
+    if (ps) {
+        if (ps->wfd > 0)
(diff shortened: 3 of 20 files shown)


grondo:
Added some build/install fixes here. `flux-imp` now gets installed to `$libexecdir/flux`, and an example `imp.toml` is added and installed to `$sysconfdir/flux/imp/conf.d`

grondo:
Actually, I removed the commit with an example `imp.toml` which is installed. Installing default config files from `make install` is probably not wise, since by default install_DATA files are always overwritten. For now we just won't install a imp.toml config (We'll run into the same problem with `sign.toml` eventually)

src/libutil/cf.c
```diff
@@ -193,6 +193,23 @@ int cf_array_size (const cf_t *cf)
     return cf ? json_array_size (cf) : 0;
```
garlick: Typo in commit message "fucntionality"

src/imp/passwd.c
```diff
@@ -0,0 +1,47 @@
...
+
+struct passwd * passwd_from_uid (uid_t uid)
+{
+    return passwd_copy (getpwuid (uid));
+}
+
+void passwd_destroy (struct passwd *pwd)
+{
+    free (pwd);
+}
```
garlick: need to free pw_name, pw_passwd

src/imp/passwd.h
```diff
@@ -0,0 +1,28 @@
...
+#ifndef HAVE_IMP_PASSWD_H
+#define HAVE_IMP_PASSWD_H 1
+
+#include <pwd.h>
+#include <sys/types.h>
+
+/*
+ *  Return a copy of the passwd entry for UID
+ *  Caller must free with passwd_destroy()
+ */
```
garlick: Perhaps mention that some fields are not copied.

grondo:
> I started a review then realized that you probably weren't ready, so I'll leave off until I get the thumbs up. Leaving my comments so far in case they are useful.

Oh, yeah sorry about the bad code. The comments are indeed helpful, we should be extra careful about code in the IMP anyway.

I'll try to make another pass and get this into mergeable shape asap.

src/imp/passwd.h
grondo: Yeah, sorry meant to update to copying all fields before pushing this up. Got too excited that it was working.. next push will have a fix.

grondo:
I pushed a bit more work here, with some added tests to ensure IMP fails as expected when conditions for operation aren't meant.

This may be ready for another look @garlick.

Still todo is to launch the job shell with a "clean" environment, instead of inheriting from the IMP (which gets the environment propagated in `flux_rexec()` by the job-exec module, cleaned up slightly by the kernel if the IMP is setuid), but I'm not sure exactly how to do this yet.

There is a well known set of FLUX_* and PATH* environment variables that are known to be required for proper job shell initialization, but I worry about propagating this knowledge directly into flux-security, and it too easily getting out-of-sync. Similar to sudo, we could have a `env-keep` and `env-delete` configuration options, but this seems a bit onerous and also prone to error.

Perhaps for now we leave the environment as-is, since the IMP can only by run by "trusted" users, and due to setuid, truly unsafe environment variables are already scrubbed.

The other missing piece is `flux-imp kill` command, which I'll be working on today.

t/t2000-imp-exec.t
```diff
@@ -0,0 +1,119 @@
...
+test_expect_success 'flux-imp exec returns error with no input' '
+	test_must_fail $flux_imp exec shell arg < /dev/null
+'
+test_expect_success 'flux-imp exec returns error with bad input ' '
+	echo foo | test_must_fail $flux_imp exec shell arg
+'
+test_expect_success 'flux-imp exec returns error with bad JSON input ' '
+	echo "{}" | test_must_fail $flux_imp exec shell arg
+'
+test_expect_success 'flux-imp exec returns error with bad JSON input ' '
```
garlick: duplicate of previous test?

t/t2000-imp-exec.t
```diff
@@ -0,0 +1,119 @@
...
+	  fake_imp_input foo | \
+	    test_must_fail $flux_imp exec echo good >badconf.log 2>&1
+	) &&
+	grep -i "Failed to load configuration" badconf.log
+'
+test_expect_success 'flux-imp exec fails with invalid stdin input' '
+	( export FLUX_IMP_CONFIG_PATTERN=sign-none.toml &&
+	  echo foo | \
+	    test_must_fail $flux_imp exec echo good >badjson.log 2>&1
+	)
```
garlick: broken && chain?   (Same for tests following this one)

src/imp/passwd.c
```diff
@@ -0,0 +1,58 @@
...
+{
+    struct passwd *pwd = NULL;
+    if (!(pwd = getpwuid (uid)))
+        return NULL;
+    return passwd_copy (pwd);
+}
+
+void passwd_destroy (struct passwd *pwd)
+{
+    free (pwd->pw_name);
```
garlick: might be good to handle NULL pwd here as its a common flux-core idiom to do this check in the destructor not the caller

src/imp/privsep.c
```diff
@@ -164,28 +164,31 @@ privsep_t * privsep_init (privsep_child_f fn, void *arg)
...
     return (status == 0 ? 0 : -1);
 }
 
+void privsep_destroy (privsep_t *ps)
+{
+    if (ps) {
+        if (ps->wfd > 0)
+            close (ps->wfd);
+        if (ps->rfd > 0)
+            close (ps->rfd);
```
garlick: I realize this is not code that originated in this PR, but since 0 is a valid file descriptor, and these values are initialized from `upfds` and `ppfds` which use -1 to designate a closed fd, should this be checking for >= 0?

t/t2000-imp-exec.t
grondo: Yeah, good catch, thanks! These were all cut-and-paste so I'll fix em all

src/imp/passwd.c
grondo: Good point!

src/imp/privsep.c
grondo: Yes, I think you're correct. I'll add a fix to this PR.

t/t2000-imp-exec.t
grondo: I think I had meant to echo some invalid JSON in the 2nd test, thanks for catching this!

grondo:
> OK, made a pass through and have a few minor comments. This is looking really good!
(I'd be inclined to approve but wip: and your comment imply that you might have more work pendiing?)

Thanks! I addressed your comments, I think, and pushed some `fixup!` commits (though the test fixes were squashed, sorry about that).

Still todo here:

 - Better handling of job shell environment
 - Add `flux-imp kill` command. (O/w jobs started under the IMP can't be killed)

If you'd like, we could merge this minimal PR and the other two could be saved for near future PRs.

garlick:
Thanks, I think this could go in and further work could be in a new PR.
Squash?

grondo:
Ok, squashed the fixup commits.

garlick:
I set merge-when-passing because there's a `.mergify.yml` in the source tree, but I had to add the label. Does that work here?

garlick:
Mergify seems stuck
 > Branch Protection is enabled and is preventing Mergify to merge the pull request. Mergify will merge when branch protection settings validate the pull request. (detail: 2 of 3 required status checks are failing.)

Should I just press the button or did you want to have a closer look at that @grondo?

grondo:
Hm, let me take a quick look. Not sure what we forgot here.

grondo:
Branch protection required that coverage checks passed. It is going to be increasingly difficult to keep coverage up unfortunately, so I've removed that protection.

garlick:
Now do we need to poke it with a stick?

grondo:
@mergifyio refresh

grondo:
@mergifyio refresh

grondo:
I had missed another codecov branch protection. Fixed now and looks like mergify worked.

garlick:
Victory!

PR #108, opened 2020-03-13
add flux_sign_wrap_as(3) to libflux-security API

grondo:
Allowing the userid to be explicitly set for `flux_sign_wrap(3)` may be useful in the development of tests in flux-framework projects that use flux-security.

For example in flux-core multiuser-exec testcases, we may want to allow job-exec to craft a fake J that encodes `FLUX_HANDLE_USERID=42` for testing. flux-core has a builtin `sign_none()` functionality that can do this, but only when not built with flux-security support.

Instead of adding a `userid` parameter to `flux_sign_wrap(3)` this PR simply renames the internal function `flux_sign_wrap_as(3)` and adds a required `userid` parameter instead of using `getuid(2)`. Then `flux_sign_wrap(3)` just calls this function directly.

This should not be merged right away, but is put up for discussion. I'll see if it can successfully be integrated with libjob in flux-core to allow optional creation of these fake sign_wrap tokens for testing.


grondo:
Thanks!

PR #109, opened 2020-03-13
imp: do not allow unprivileged exec by default 

grondo:
It is giving me some heebie-jeebies that a system instance install of flux-core with the current flux-security where the IMP is *not* installed setuid will allow any user to run anything as user `flux`.

This PR cleans that up by having the IMP refuse to exec anything by default when installed unprivileged. Since the unprivileged exec is used in testing, a new config parameter `exec.allow-unprivileged-exec` is added.


grondo:
Thanks!

PR #110, opened 2020-03-17
imp: add a kill command

grondo:
This PR adds a `kill` command to the IMP, used as `flux-imp kill SIGNAL PID`. If the user running the IMP is in the `exec.allowed-users` config, and the IMP is installed setuid, then this command allows the requesting user to kill processes running as other users as long as the process is running in a cgroup owned by the requesting user.

The current version assumes that the multiuser Flux has been started by systemd with `Delegate=yes`.

I'm not exactly sure how to do any type of testing for this new code. I think within travis-ci builders, we might find that the cgroup filesystem isn't even mounted.

I will at least try to push some sanity checks for the failure modes of the `kill` subcommand -- user not in allowed-users, etc.

### DIFF
--- a/src/imp/kill.c
+++ b/src/imp/kill.c
@@ -1,0 +1,236 @@
+/************************************************************\
+ * Copyright 2020 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+/* flux-imp kill - signal tasks on behalf of requestor when authorized
+ *
+ * PURPOSE:
+ *
+ *  Allow a non-privileged Flux instance to signal processes in
+ *  jobs running as different users.
+ *
+ * OPERATION:
+ *
+ *  The IMP kill command currently works under the assumption that the
+ *  multiuser instance is running under systemd with Delegate=yes. This
+ *  setting directs systemd to delegate ownership of the flux.service
+ *  cgroup to the user under which Flux is runnng, e.g. "flux".
+ *
+ *  Since all jobs will executed within this cgroup or a child,
+ *  flux-imp kill may authorize signal delivery to any task where
+ *  the tasks cgroup is owned by the requesting user.
+ *
+ */
+
+#if HAVE_CONFIG_H
+#include <config.h>
+#endif
+
+#include <stdio.h>
+#include <unistd.h>
+#include <stdlib.h>
+#include <stdint.h>
+#include <string.h>
+#include <errno.h>
+
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <pwd.h>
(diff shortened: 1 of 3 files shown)


garlick:
Excellent!  I'll do a quick manual test.

grondo:
I'll be pushing an update to flux-framework/flux-core#2822 with changes to `job-exec` to use `flux-imp kill` as well. It won't pass Travis until we tag a new release of flux-security that can be pulled into the docker image.

garlick:
Oh right, I'll grab that too for my test.

garlick:
Looks good:

Cancel my own job:
```
$ flux mini submit sleep 3600
1477508293722112
$ flux jobs
             JOBID USER     NAME       STATE    NTASKS NNODES  RUNTIME RANKS
  1477508293722112 garlick  sleep      RUN           1      1   1.332s 0
$ flux job cancelall -f
flux-job: Canceled 1 jobs (0 errors)
$ flux jobs
             JOBID USER     NAME       STATE    NTASKS NNODES  RUNTIME RANKS
$ flux job attach 1477508293722112
21.386s: job.exception type=cancel severity=0
```

Try to cancel someone else's job
```
$ sudo -u flux flux mini submit sleep 3600
1480435565592576
$ flux job cancel 1480435565592576
flux-job: 1480435565592576: guests can only raise exceptions on their own jobs
$ sudo -u flux flux job cancel 1480435565592576
$ flux job attach 1480435565592576
flux-job: flux_job_event_watch_get: Operation not permitted
$ sudo -u flux flux job attach 1480435565592576
61.399s: job.exception type=cancel severity=0
flux-job: task(s) exited with exit code 143
```

grondo:
> $ flux job attach 1480435565592576
> flux-job: flux_job_event_watch_get: Operation not permitted

We should fix some of these error messages to be more user-friendly.

src/imp/kill.c
```diff
@@ -0,0 +1,236 @@
...
+    return false;
+}
+
+static void check_and_kill_process (struct imp_state *imp, pid_t pid, int sig)
+{
+    uid_t user = getuid ();
+    struct pid_info *p = NULL;
+
+    if (!imp_kill_allowed (imp->conf))
+        imp_die (1, "kill command not allowed");
```
garlick: This might be an uninformed question:  does this code need an explicit check to avoid signaling root processes like the exec subcommand has?

grondo: Hm, it shouldn't come up unless the instance is able to start a root process. If the cgroup in which the process is running is not owned by current uid, then `imp kill` will refuse to kill it.

Additionally, the actual userid of processes isn't checked at this point.

grondo:
> Also (minor) there are a few spaces on the end of lines.

Hm, sorry about that. I'll force-push a fix.

src/imp/kill.c
garlick: Sounds good to me, thanks,

grondo:
Oops, I was still working on a test. I'll create a separate PR for that.

garlick:
Oh sorry!  I had the wrong impression.

PR #111, opened 2020-03-18
imp: add tests for flux-imp kill, fix discovered issues

grondo:
This PR adds the `flux-imp kill` tests promised in #110. 

Additionally a memleak fix found along the way, and for testing, allow non-root users running the IMP to kill their own processes. (Still pondering if this one should require an option.)

### DIFF
--- a/src/imp/kill.c
+++ b/src/imp/kill.c
@@ -53,6 +53,7 @@
 
 struct pid_info {
     pid_t pid;
+    uid_t pid_owner;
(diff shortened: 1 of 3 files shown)


grondo:
If this goes in, we may want to create a flux-security tag so that flux-core test docker images can be built against the latest flux-security code.

Alternately, we could update the flux-core docker build to allow builds against flux-security@master.

grondo:
Thanks!

PR #112, opened 2020-03-19
Add release notes for flux-security v0.4

grondo:
Update NEWS.md for flux-security v0.4.

With the basic flux-imp `exec` and `kill` "working" for some definition thereof, now might be a good time to tag a flux-security release, so we can update the version built in flux-core test docker images.


grondo:
Thanks!

PR #113, opened 2020-07-09
libtomlc99: Rename blacklist to blocklist

chu11:


grondo:
Thanks! Want to push this upstream, or has it already been done?

chu11:
upstream as in `flux-core`?  Got a PR for `flux-core` as well.  https://github.com/flux-framework/flux-core/pull/3040#issuecomment-656157803  `tomlc99`'s project doesn't seem to have the term in their repo, I assume they removed it.

grondo:
No, upstream as in the source repo for `libtomlc99`. We didn't write that code and vendored it from somewhere. I assume it is listed in the source or history of commits.

grondo:
> tomlc99's project doesn't seem to have the term in their repo, I assume they removed it.

Ah, then why don't we cherry pick those commits, or update to latest tomlc99?

chu11:
Hmmm. I can't find this test file anywhere including the `tomlc99` history.  From logs:

```
    TAP test replaces upstream test.  It runs the BurntSushi test
    inputs through the parser and checks for expected pass/fail.
    See also: https://github.com/BurntSushi/toml-test
```

@trws did you write this test file?  I can't find it in `toml-test` either.

chu11:
I realized `blocklisted` probably isn't a word, so in one place, `blacklisted` -> `blocklist`

grondo:
Actually, probably @garlick wrote those tests (`git blame` is your friend) or pulled those down from the BurntSushi repo. No upstream changes required. Sorry about the noise!

garlick:
Yeah I'm to blame for the TAP test.  Here's the commit that added it, which explains the BurntSushi tie-in:
```
commit f35927bec7b450e4e4e0c7254fae525e09005212
Author: Jim Garlick <garlick.jim@gmail.com>
Date:   Thu Nov 16 15:25:17 2017 -0800

    libtomlc99: add simple TAP test
    
    The tomlc99 project includes a copy of the general TOML
    test suite: https://github.com/BurntSushi/toml-test
    
    In order to avoid taking on its Golang dependency,
    we simply add a small TAP test that runs all the valid and
    invalid TOML input files from the BurntSushi suite through
    the libtomlc99 parser and ensure that they pass/fail as expected.
    
    A few invalid input files are not handled properly.  Add a
    table of them with the reason for failure so the TAP test
    can skip them and emit the reason for skipping.
```

grondo:
Thanks, @chu11!

PR #115, opened 2020-11-04
update mergify config for move to travis-ci.com

grondo:
flux-security has been migrated from travis-ci.org to travis-ci.com.

I think this is the one required change due to the migration.


grondo:
Thanks!

PR #121, opened 2021-07-14
ci: use github actions to run CI

grondo:
flux-security was still tied to Travis-CI for CI builds. This PR switches to Github Actions like other flux-framework projects.

This PR mostly copies the infrastructure for docker-based CI builds in Github Actions from flux-core, with modifications required for the flux-security project. Since the `fluxrm/testenv` images already contain the build requirements for flux-security (since a build of flux-security is currently the first step in setting up those containers), it is convenient to go ahead and use those same images for flux-security CI.

This means we can expand the distros on which automated tests are run to the same set as used with flux-core.

One package was missing from the fedora33 image (libubsan). That is installed by hand here for now, but maybe I'll update our testenv image and drop that addition here if there is time.


grondo:
Just pushed an additional update to the `.mergify.yml` to remove the required Travis-CI status checks and add the github actions status checks. This PR will now need to be manually merged.

SteVwonder (approved):
Thanks @grondo! LGTM!

grondo:
Thanks, @SteVwonder! Mind pressing the merge button? This one needs to be merged manually due to changes to `.github/workflows`

grondo:
Thanks!

Issue #101, opened 2020-02-28
libutil: cf: need "paranoid" mode


grondo:
The current config parsing wrapper code opens and parses TOML from files without any permissions checking. For security-significant tools like the IMP, there should be a paranoid mode where reading the configuration file fails if permissions are too permissive or the file is owned by the wrong user.

PR #122, opened 2021-07-15
check file permissions and ownership of configuration when running privileged code

grondo:
As described in #101, TOML config files opened by the IMP should be checked for secure permissions and ownership. This PR adds a "paranoid" mode to `libutil/cf.c` which is enabled by default whenever the effective UID of the process is `0` or does not match the real UID (setuid mode). This mode can also be enabled by explicitly setting `enable-path-paranoia = true` in a `cf_t` object before calling one of the functions that reads from files (e.g. `cf_update_glob()`, `cf_update_file()`).

Paranoid mode can also be explicitly _disabled_ by setting `disable-path-paranoia = true`. This is only used (and only designed) for testing, where it is convenient to test the IMP running in sharness tests when reading from config files in the sharness trash directory with regular user ownership. Therefore, you will see that the test-only IMP `src/imp/flux-imp` is built with this setting enabled.

In order to allow setting these special configuration keys, a new `cf_update_pack(3)` function was introduced, which allows a config object to be created, some initial keys set, then configuration read from files. This is used to disable path-paranoia in the test IMP.

This concept also had to be extended to `flux_security_configure(3)`, which does not expose the `cf_t` interface. Therefore, two new flags were added which can be used with `flux_security_create(3)`:
```
enum {
    FLUX_SECURITY_DISABLE_PATH_PARANOIA     = 0x1,
    FLUX_SECURITY_FORCE_PATH_PARANOIA       = 0x2,
};
```

Only one of these flags may be used at a time. The test IMP, for example, uses `FLUX_SECURITY_DISABLE_PATH_PARANOIA` -- again, this flag should only be used in testing. Forcing path security checks, however, may actually be useful when a normal user process that is verifying signatures opens the config (root or setuid processes would get security checks by default)

You might be wondering why secure file and directory checks can't be the default. The problem is that we have several types of processes reading different types of configuration. For the IMP and perhaps (if required) the _system_ Flux instance, the reading process should ensure its configuration is secure. However, normal users should be able to use any signing or other config files they would like, whether they be system owned, or personal copies. Therefore it doesn't make sense to check the file and directory ownership and permissions in that case (there are just too many corner cases)

This is built on top of #121 to get the improved CI.

### DIFF
--- a/src/lib/context.c
+++ b/src/lib/context.c
@@ -25,6 +25,7 @@
 
 struct flux_security {
     cf_t *config;
+    int flags;
     struct aux_item *aux;
     char error[200];
     int errnum;
@@ -50,16 +51,31 @@ void security_error (flux_security_t *ctx, const char *fmt, ...)
     }
 }
 
+static bool valid_flags (int flags)
+{
+    /*  Currently valid flags:
+     *   Only one of FLUX_SECURITY_DISABLE_PATH_PARANOIA or
+     *   FLUX_SECURITY_FORCE_PATH_PARANOIA may be set at a time.
+     *   O/w, flags must be unset.
+     */
+    if (flags == 0
+        || flags == FLUX_SECURITY_DISABLE_PATH_PARANOIA
+        || flags == FLUX_SECURITY_FORCE_PATH_PARANOIA)
+        return true;
+    return false;
+}
+
 flux_security_t *flux_security_create (int flags)
 {
     flux_security_t *ctx;
 
-    if (flags != 0) { // not used yet
+    if (!valid_flags (flags)) {
         errno = EINVAL;
         return NULL;
     }
     if (!(ctx = calloc (1, sizeof (*ctx))))
         return NULL;
+    ctx->flags = flags;
     return ctx;
 }
 
@@ -82,7 +98,6 @@ int flux_security_last_errnum (flux_security_t *ctx)
     return ctx ? ctx->errnum : 0;
 }
 
-
 int flux_security_configure (flux_security_t *ctx, const char *pattern)
 {
     struct cf_error cfe;
@@ -98,6 +113,23 @@ int flux_security_configure (flux_security_t *ctx, const char *pattern)
     if (!(cf = cf_create ())) {
         security_error (ctx, NULL);
         return -1;
+    }
+    if (((ctx->flags & FLUX_SECURITY_DISABLE_PATH_PARANOIA)
+        && cf_update_pack (cf,
+                           &cfe,
+                           "{s:b}",
+                           "disable-path-paranoia", true) < 0)
+        || ((ctx->flags & FLUX_SECURITY_FORCE_PATH_PARANOIA)
+            && cf_update_pack (cf,
+                               &cfe,
+                               "{s:b}",
+                               "enable-path-paranoia", true) < 0)) {
+        security_error (ctx,
+                        "%s: failed to apply ctx flags: %s",
+                        pattern,
+                        cfe.errbuf);
+        goto error;
+
     }
     if ((n = cf_update_glob (cf, pattern, &cfe)) < 0) {
         security_error (ctx, "%s::%d: %s",
--- a/src/libutil/Makefile.am
+++ b/src/libutil/Makefile.am
@@ -30,7 +30,9 @@ libutil_la_SOURCES = \
 	sha256.h \
 	macros.h \
 	aux.c \
-	aux.h
+	aux.h \
+	strlcpy.c \
+	strlcpy.h
 
 TESTS = \
 	test_hash.t \
--- a/src/libutil/cf.c
+++ b/src/libutil/cf.c
@@ -268,6 +268,44 @@ int cf_update_file (cf_t *cf, const char *filename, struct cf_error *error)
     return update_object (cf, filename, NULL, 0, error);
 }
 
+int cf_update_pack (cf_t *cf, struct cf_error *error, const char *fmt, ...)
+{
+    json_error_t err;
+    json_t *o;
+    va_list ap;
+    int saved_errno;
+    int rc;
+
+    if (!cf || json_typeof ((json_t *)cf) != JSON_OBJECT) {
+        errprintf (error, NULL, -1, "invalid config object");
+        errno = EINVAL;
+        return -1;
+    }
+    if (!fmt) {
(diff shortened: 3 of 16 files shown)


grondo:
Actually set `WIP` for now since I haven't checked this on a real system instance yet.

grondo:
I double-checked the new code in a system instance configuration and verified operation, so I've removed WIP.
I also made the PR description more generic since the checks apply to more than just the IMP (thought that is the most important visible change)

grondo:
rebased

src/libutil/Makefile.am
```diff
@@ -30,7 +30,9 @@ libutil_la_SOURCES = \
 	sha256.h \
 	macros.h \
 	aux.c \
-	aux.h
+	aux.h \
+	strlcpy.c \
+	strlcpy.h
```
garlick, line 35: commit message for 1492474f8df8524d9c2dcead81a3d87e0c043367: since MUNGE `path.[ch]` is no longer being added in this PR, justification should be updated.

"because we likesssssssss it?"

src/libutil/cf.c
```diff
@@ -210,6 +215,121 @@ bool cf_array_contains (const cf_t *cf, const char *str)
...
+    }
+    if (st.st_gid != 0
+        && st.st_gid != getegid ()
+        && (st.st_mode & S_IWGRP)
+        && !(st.st_mode & S_ISVTX)) {
+        errprintf (error, filename, -1,
+                   "parent directory is group-writeable without sticky bit");
+        errno = EINVAL;
+        return false;
+    }
```
garlick, line 257: So sys admins would need to ensure that config tree for flux-security is all `root:root`.  `flux:flux` would not be appropriate since flux isn't supposed to have any special privileges.  Now that I state that out loud it seems obvious... Duh.

The other allowed case would be for testing where some user owns everything and runs without setuid.  Got it.

src/libutil/cf.c
```diff
@@ -210,6 +215,121 @@ bool cf_array_contains (const cf_t *cf, const char *str)
...
+    if (!S_ISREG (st.st_mode)) {
+        errprintf (error, filename, -1, "File is not a regular file");
+        errno = EINVAL;
+        return false;
+    }
+    if (is_symlink) {
+        errprintf (error, filename, -1, "Refusing to open symbolic link");
+        errno = EINVAL;
+        return false;
+    }
```
garlick, line 296: I take it if it's a symlink to a special file, you wanted to call that out rather than just bail on a symlink no matter what it points to?

Edit: oh when I came to the unit tests I now see why this makes things simpler.

src/lib/context.c
```diff
@@ -99,6 +114,24 @@ int flux_security_configure (flux_security_t *ctx, const char *pattern)
         security_error (ctx, NULL);
         return -1;
     }
+    fprintf (stderr, "ctx->flags = 0x%x\n", ctx->flags);
```
garlick, line 117: Loose `fprintf()` from debugging?

garlick (approved):
This looks good to me.

dun (approved):
Looks good to me! :+1:

src/libutil/Makefile.am
grondo: How about this? 
```
    libutil: add strlcpy
    
    Problem: strncpy(3) is a frequent source of programmer mistakes
    which directly result in security vulnerabilities. In order to be
    used safely, strncpy(3) requires an initialized buffer, or manual NUL
    termination, plus manual checks for length to ensure a result buffer
    without NUL termination is not created. This should not have to be
    the case in a project where the purpose is to create secure code.
    
    strlcpy(3), on the other hand, is designed "to be a safer, more
    consistent, and less error prone replacement" for strncpy(3) [1]. With
    its guarantee of NUL termination and safer semantics, it behooves us
    to include this implementation in the flux-security project.
    
    Note that since strlcpy(3) is known not be included in glibc, we can
    include it wholesale here. However, when porting to BSD systems
    presence of strlcpy() in libc may need to be detected and this
    source file excluded from the build if an implementation is found.
    
    [1] https://man.openbsd.org/strlcpy
```

grondo: I'm also fine with switching back to `strncpy(3)` if you'd prefer. It is possible to safely use with proper length checks and initialization, and it is only used a handful of times currently. Perhaps our use of LGTM and ASan/UBSan and future use of fuzzing (if we can get that working) will make mistakes less likely.

src/libutil/cf.c
grondo: Yeah, and TBH I just took the example from munged keyfile tests. Perhaps @dun could comment on what the best approach is here. In the case of flux-security config files, it may actually be convenient to allow symlinks as long as the target also has the correct permissions/ownership?

src/lib/context.c
grondo: Thanks for catching that!

src/libutil/cf.c
dun: munged checks the permissions and ownership of its directories all the way up to the root. Since a symlink could point outside of this directory subtree, munged simply disallows symlinks. See:

https://cwe.mitre.org/data/definitions/61.html
https://cwe.mitre.org/data/definitions/59.html

If you're going to allow symlinks, I'd recommend the symlink target should have the same ownership and permission checks as the original path (i.e., you'd need to, at least, also check the ownership and permissions of the target directory).

src/libutil/Makefile.am
garlick: Wow nice commit message!  I say let's keep it.

src/libutil/cf.c
grondo: Thanks for the background @dun! That is really helpful. I think at this point I'd like to err on the side of being overly cautious and just disallow symlinks when ownership and permission checks are enabled. If this becomes onerous in the future, then this requirement could potentially be relaxed.

Another mitigating factor here is that the cf file and permission checks are only enabled if the process is running with elevated privileges or setuid (euid != uid), so normal users should be able to use symlinks to their heart's content. :heart_eyes: :link:

grondo:
I've force pushed an update to this PR that addresses comments above (mainly: rewording commit that adds `strlcpy()`)

garlick:
Looks good!  Already approved.  Do we have merge when passing here or do you need a button pusher?

grondo:
Thanks! We should have mergify set up here so I've set MWP

PR #123, opened 2021-08-31
imp: add run subcommand

### LINKED ISSUE #120: Support imp exec of prolog/epilog
The IMP needs support for execution of privileged prolog and epilog scripts in the system instance. This will likley be implemented as `imp prolog` and `imp epilog` subcommands, with configuration to allow access only to a designated user (e.g. `flux`). The explicit path to prolog and epilog would be part of the IMP configuration, not selectable at runtime. A subset of environment variables may be passed through, e.g. to indicate the jobid, user, etc for which the prolog/epilog is running, and any environment variables that may explicitly influence optional behavior of these scripts.

grondo:
As described in #120, the IMP will need to assist the Flux system instance with execution of job prolog and epilog scripts, which typically require elevated privileges.

Instead of adding separate `flux-imp prolog` and `flux-imp epilog` subcommands, this PR adds a new `run` subcommand to the IMP which is invoked like `flux-imp run COMMAND`. Commands to execute under the IMP `run` subcommand are added to the IMP configuration under a `run` table by name, e.g. `[run.prolog]` and `[run.epilog]`. Required keys in these run command tables include:

  * `path` - the path to the executable to run
  *  `allowed-users` - an array of usernames that are allowed to run this command, e.g. `[ "flux" ]`
  *  `allowed-environment` - an array of environment variables and/or `glob(7)` patterns which match environment variables that the IMP will propagate to the command. By default only `FLUX_JOB_ID` and `FLUX_JOB_USERID` are propagated.

(Edit: actually `allowed-environment` key is not required)

The only other environment variables set in the executed command include:

 - `FLUX_OWNER_USERID` - the UID which called `flux-imp run COMMAND`
 - `HOME` and `USER` -  set appropriately to the privileged user running `COMMAND` (since these variables are often assumed to be present in many scripts)
 - `PATH` is set explicitly to `/usr/sbin:/usr/bin:/sbin:/bin`

This will allow the Flux system instance, when properly configured, to execute `flux-imp run prolog` and `flux-imp run epilog`, leaving a facility in place to give Flux the ability to execute other work with root privileges later (e.g. `flux-imp run healthcheck` or similar)

Fixes #120

### DIFF
--- a/src/imp/run.c
+++ b/src/imp/run.c
@@ -64,6 +64,10 @@
 
 extern char **environ;
 
+#if CODE_COVERAGE_ENABLED
+extern void __gcov_flush ();
+#endif
+
 static const cf_t *imp_run_lookup (struct imp_state *imp,
                                    const char *name)
 {
@@ -178,6 +182,9 @@ imp_run (const char *name,
 
     args[0] = path;
     args[1] = NULL;
+#if CODE_COVERAGE_ENABLED
+    __gcov_flush ();
+#endif
     execve (path, (char **) args, env);
 
     if (errno == EPERM || errno == EACCES)
--- a/src/imp/run.c
+++ b/src/imp/run.c
@@ -1,0 +1,338 @@
+/************************************************************\
+ * Copyright 2021 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+/* flux-imp run
+ *
+ * PURPOSE:
+ *
+ *  Allow a non-privileged Flux instance to execute a configured
+ *   named executable in the [run] section (e.g. "prolog" or "epilog")
+ *
+ * OPERATION
+ *
+ *  Unprivileged child reads "command" requested to execute on command line,
+ *   and any allowed environment variables and sends kv struct to parent
+ *
+ *  Lookup configuration for "command"
+ *    - ensure calling user is allowed to execute "command"
+ *    - set absolute path for command
+ *    - capture allowed environment variables based on allowed-environment
+ *
+ *  Environment variables set by IMP:
+ *  - FLUX_OWNER_USERID - uid of the user running the IMP
+ *  - PATH - "/usr/sbin:/usr/bin:/sbin:/bin"
+ *  - HOME
+ *  - USER
+ *
+ *  - switch uid and gid to effective uid gid
+ *
+ *  - wait for child to terminate successfully
+ *
+ *  - execute command
+ */
+
+#if HAVE_CONFIG_H
+#include <config.h>
+#endif
+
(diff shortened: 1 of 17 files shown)


src/imp/run.c
```diff
@@ -0,0 +1,333 @@
...
+
+    if (kv_expand_environ (kv_env, &env) < 0)
+        imp_die (1, "Unable to set %s environment", name);
+
+    if (chdir ("/") < 0)
+        imp_die (1, "run: failed to chdir to /");
+
+    args[0] = path;
+    args[1] = NULL;
+    execvpe (path, (char **) args, env);
```
garlick, line 183: `execvpe()` searches for executable in PATH.  Since we're ensuring the path is not relative, would `execve()` be more appropriate?

src/imp/run.c
```diff
@@ -0,0 +1,333 @@
...
+    char **env;
+    const char *args[2];
+    int exit_code;
+
+    if (!(path = cf_string (cf_get_in (run_cf, "path")))
+        || path[0] != '/')
+        imp_die (1, "run: %s: invalid path", name);
+
+    if (path[0] != '/')
+        imp_die (1, "run %s: relative path not allowed", name);
```
garlick, line 154: Wondering if any other checks are needed here, such as verifying that the executable or the path containing it are not writable by others?

garlick (commented):
Nicely done!

Couple of comments/questions on a first review pass.

src/imp/run.c
grondo: Yeah, not sure how I ended up with `execvpe()` there :shrug: It should definitely be `execve()`. Thanks for catching that.

src/imp/run.c
grondo: Yeah, good thought. Too bad the functions to check file and directory ownership and permissions are private to `libutil/cf.c` -- that wasn't very foresighted. :disappointed: 

We'll have to maybe split out those functions in libutil...

src/libutil/kv.c
```diff
@@ -500,6 +500,44 @@ struct kv *kv_split (const struct kv *kv1, const char *prefix)
...
+    if (kv == NULL || envp == NULL) {
+        errno = EINVAL;
+        return -1;
+    }
+
+    while ((var = kv_next (kv, var)))
+        size++;
+    if (!(*envp = calloc (size+1, sizeof (char *))))
+        return -1;
+    env = *envp;
```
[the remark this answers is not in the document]
grondo: Good catch, thanks!

grondo:
Thanks for the comments @garlick! I've force-pushed an update to this PR to address some of the comments.
Some of the minor work was squashed (updates to `kv_expand_environ()` and switch to `execve`).

The other major work was to split out `path_is_secure()` into `src/libutil/path.[ch]` so it could be used directly from within the IMP `run` command. Associated fixup commits to check the target path security and updates to the tests are left tacked on the end for now.

garlick (approved):
This looks great!  I couldn't find anything less than perfect here.  Set MWP at will!

grondo:
Thanks! I've squashed the fixup commits and will set MWP.

PR #125, opened 2021-09-08
ci: add fedora 34 and -fanalyzer build to ci checks

grondo:
This PR updates the flux-security automated CI build to build on Fedora Core 34. The address sanitizer build is updated to fedora 34, and to the normal build we add the `-fanalyzer` flag to `CFLAGS` which turns on gcc 11.2's build time static analysis.

This only found a couple issues (fixed here), but it will be nice to have this enabled for PRs going forward.


src/libca/sigcert.c
```diff
@@ -368,15 +368,19 @@ int sigcert_store (const struct sigcert *cert, const char *name)
         goto error;
     if (sigcert_fwrite_public (cert, fp) < 0)
         goto error;
-    if (fclose (fp) < 0)
```
garlick, line 371: Good catch, `-fanalyer`!

There's a small typo in the commit message: "failse"

garlick (approved):
Nice, just once small commit message typo noticed.

src/libca/sigcert.c
grondo: Thanks! Fixed and pushed.

grondo:
Setting MWP.

grondo:
I forgot to update the `.mergify.yml`. This will now need to be manually merged I think.

Issue #127, opened 2022-01-27
imp: allow exec to take more than one argument to COMMAND


grondo:
Problem: Currently the `flux-imp exec` functionality is limited to supplying a single argument to the `flux-shell` (assumed to be the JOBID for which the shell is being invoked). However, a system instance may want to supply extra options to the job shell via a new `job-shell-options` configuration parameter. For example, a new `--reconnect` option to tell the shell to attempt to reconnect to the broker if the current connection is lost, or an `--initrc` option to direct the system instance shell to a different initrc.

The IMP should allow these extra options to be passed on the command line by the exec system.

PR #128, opened 2022-01-27
imp: support multiple arguments to job shell in exec

grondo:
This PR addresses #127.

Instead of supporting only 1 argument, all arguments passed in the call the `flux-imp exec` are passed along to the job shell. This will allow the system instance to add extra options to the job shell based on configuration, e.g. a `--reconnect` option.

### DIFF
--- a/src/imp/exec/exec.c
+++ b/src/imp/exec/exec.c
@@ -54,7 +54,7 @@ struct imp_exec {
 
     const char *J;
     const char *shell;
-    const char *arg;
+    struct kv *args;
     const void *spec;
     int specsz;
 };
@@ -97,6 +97,7 @@ static void imp_exec_destroy (struct imp_exec *exec)
         flux_security_destroy (exec->sec);
         json_decref (exec->input);
         passwd_destroy (exec->imp_pwd);
+        kv_destroy (exec->args);
         free (exec);
     }
 }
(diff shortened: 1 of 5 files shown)


grondo:
Rebased. It is going to be difficult to get coverage any higher since all uncovered paths are unlikely error handling.

garlick (approved):
LGTM.

Thinking this through out loud: if the IMP is not locked down so that users other than `flux` can run it, is there anything they can do by specifying weird options here?  The shell path and allowed signing mechs are locked down by configuration, and the user transition is checked against the J signature.  So the best someone up to mischief could do is maybe tell the IMP to run the installed flux-shell as themselves with arbitrary arguments.  Yeah, sounds good to me!

grondo:
Yeah, good question.

I think there actually could be a bit of a problem when the IMP isn't locked down to a system instance user. As the IMP includes configuration for allowed-shells it should probably also include a list of allowed options. For example, in the eventual case where a non system user is allowed to execute the IMP to run other user's jobs, they should not be able to supply the `--initrc` option, which would essentially allow the calling user to select arbitrary code as the user that signed _J_. At this point probably the only allowed option should be `--reconnect`, so maybe the appropriate thing to do is to make this an option to `flux-imp exec` instead of allowing arbitrary options to be passed through the IMP to the shell?

garlick:
Up to you, but it seems fairly safe for now, since we don't yet support multi-user instances started by arbitrary users, and users signed job requests are protected from access in the system instance.  An attacker would first need to get the user to sign a request and give it to them, then apply it as you described.  So maybe opening an issue would be sufficient for us to move forward now with this PR as is?

grondo:
Ok, that is reasonable. The calling user would not only require permission to execute the IMP, but would also have to be explicitly listed in `allowed-users` _and_ have access to a signed _J_ from another user.

I'll open an issue and set MWP here. Thanks!

PR #129, opened 2022-01-28
mergify:  replace strict merge with queue+rebase

grondo:
Problem: Mergify.io has deprecated the strict merge mode action:

 https://blog.mergify.io/strict-mode-deprecation/

Replace the strict mode configuration with a default queue with a
rebase action which should be equivalent.


grondo:
I think this will need a manual merge since it updates the mergify config.

garlick (approved):
Thanks!

PR #134, opened 2022-01-31
ci: rename centos7/8 to el7/8

grondo:
Problem: Upstream docker images have been renamed from centos7 and
centos8 to el7 and el8 to reflect the fact that CentOS is now EOL.

Update image names used in CI accordingly.

`fluxrm/testenv:el7` and `fluxrm/testenv:el8` have already been pushed to dockerhub, so this should be safe to merge now. (i.e. flux-security doesn't depend on flux-core images)


chu11 (approved):

grondo:
This will need a manual merge since mergify config changed

grondo:
Thanks!

PR #135, opened 2022-02-01
fix auto-deploy of flux-security release on tag

grondo:
The auto-deploy actions for the last flux-security tag didn't work.

I think I've run down the problem (of course, you can't check without another tag), with fixes in this simple PR. Since the workflow is modified, this may require a manual merge. Hopefully things just work next time.


garlick (approved):
Thanks!

PR #136, opened 2022-02-01
testsuite: fix spell checker and some spelling errors

garlick:
Apparently our `spellcheck` script here and in flux-core has not been working for a while, since `IPv4` was added to the dictionary.  Fix a bug in the `spellcheck` script and then several spelling errors that emerged.

Also, add a libsodium "resource" to `flux-config-security-sign(5)`.

Will submit a PR to fix flux-core as well.


grondo (approved):
Ugh! Good catch!

One typo in commit message fe22a7a `os it fails` :arrow_right: `as it fails`

garlick:
Sheesh sorry for my typos.  I had already re-pushed twice to fix other typos.  Fixed that one too, and over in flux-core.

Hence the need for spell checking!!!

PR #139, opened 2022-02-08
doc: fix manpage typo

chu11:
Problem: The flux-config-security manpage was typoed as
flux-security-config.

Fix the typo.


garlick (approved):
Oops, good catch!

PR #142, opened 2022-06-04
imp: fix kill functionality with unified cgroups hierarchy

grondo:
This PR allows the IMP `kill` functionality to work on distros and configurations where unified cgroups are in use.

A simple function is added which determines whether unified cgroups are active, and the mount point for the cgroups hierarchy, using the rules described at https://systemd.io/CGROUP_DELEGATION/. Otherwise, changes are minimal. 

I've tested this out on a local Ubuntu 22.04 system, but it would be good if @garlick could test this on his raspbian system as well.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -118,6 +118,9 @@ LT_INIT
 #  Checks for header files
 #
 AC_HEADER_STDC
+AC_CHECK_HEADERS( \
+  [linux/magic.h] \
+)
 
 #
 #  Checks for packages
--- a/src/imp/kill.c
+++ b/src/imp/kill.c
@@ -45,10 +45,16 @@
 
 #include <sys/types.h>
 #include <sys/stat.h>
+#include <sys/vfs.h>
+#ifdef HAVE_LINUX_MAGIC_H
+#include <linux/magic.h>
+#endif
+
 #include <pwd.h>
 #include <signal.h>
 
 #include "src/libutil/kv.h"
+#include "src/libutil/strlcpy.h"
 
 #include "imp_log.h"
 #include "imp_state.h"
@@ -62,10 +68,53 @@ struct pid_info {
     uid_t cg_owner;
 };
 
-/*  Hard-coded path to systemd cgroup mount directory. This may
- *   need to be moved to configuration at some point.
+struct cgroup_info {
+    char mount_dir[PATH_MAX + 1];
+    bool unified;
+};
+
+/*  Determine if this system is using the unified (v2) or legacy (v1)
+ *   cgroups hierarchy (See https://systemd.io/CGROUP_DELEGATION/)
+ *   and mount point for systemd managed cgroups.
  */
-static const char cgroup_mount_dir[] = "/sys/fs/cgroup/systemd";
+static int cgroup_info_init (struct cgroup_info *cg)
+{
+    struct statfs fs;
+
+    (void) strlcpy (cg->mount_dir, "/sys/fs/cgroup", sizeof (cg->mount_dir));
+    cg->unified = true;
+
+    if (statfs (cg->mount_dir, &fs) < 0)
+        return -1;
+
+#ifdef CGROUP2_SUPER_MAGIC
+    /* if cgroup2 fs mounted: unified hierarchy for all users of cgroupfs
+     */
+    if (fs.f_type == CGROUP2_SUPER_MAGIC)
+        return 0;
+#endif /* CGROUP2_SUPER_MAGIC */
+
+    /*  O/w, if /sys/fs/cgroup is mounted as tmpfs, we need to check
+     *   for /sys/fs/cgroup/systemd mounted as cgroupfs (legacy).
+     *   We do not support hybrid mode (/sys/fs/cgroup/systemd or
+     *   /sys/fs/cgroup/unified mounted as cgroup2fs), since there were
+     *   no systems on which to test this configuration.
+     */
+    if (fs.f_type == TMPFS_MAGIC) {
+
+        (void) strlcpy (cg->mount_dir,
+                        "/sys/fs/cgroup/systemd",
+                        sizeof (cg->mount_dir));
+        if (statfs (cg->mount_dir, &fs) == 0
+            && fs.f_type == CGROUP_SUPER_MAGIC) {
+            cg->unified = false;
+            return 0;
+            }
+    }
+
+    /*  Unable to determine cgroup mount point and/or unified vs legacy */
+    return -1;
+}
 
 /*  Return the systemd cgroup path for PID `pid` in the provided buffer
  *  Looks up the 'name=systemd'[*] subsystem relative cgroup path in
@@ -82,6 +131,10 @@ static int pid_systemd_cgroup_path (pid_t pid, char *buf, int len)
     int n;
     char file [4096];
     char *line = NULL;
+    struct cgroup_info cgroup;
+
+    if (cgroup_info_init (&cgroup) < 0)
+        return -1;
 
     n = snprintf (file, sizeof(file), "/proc/%ju/cgroup", (uintmax_t) pid);
     if ((n < 0) || (n >= (int) sizeof(file))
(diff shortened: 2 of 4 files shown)


garlick:
This appears to kill the shell
```
 garlick@picl0:~$ flux mini run -t 10s -N1 stress --cpu 4
stress: info: [3053] dispatching hogs: 4 cpu, 0 io, 0 vm, 0 hdd
10.459s: job.exception type=timeout severity=0 resource allocation expired
flux-job: task(s) Killed

 garlick@picl0:~$ sudo flux dmesg
2022-06-04T05:17:49.132577Z job-exec.info[0]: job-exception: id=4818363150761984: resource allocation expired
2022-06-04T05:17:49.132819Z job-exec.debug[0]: exec_kill: 4818363150761984: signal 14
2022-06-04T05:17:54.138025Z job-exec.debug[0]: Sending SIGKILL to job 4818363150761984
2022-06-04T05:17:54.138231Z job-exec.debug[0]: exec_kill: 4818363150761984: signal 9
2022-06-04T05:17:54.681828Z sched-simple.debug[0]: free: rank2/core[0-3]

```
but leaves tasks running
```
 garlick@picl2:~$ ps -fu garlick
UID          PID    PPID  C STIME TTY          TIME CMD
garlick     3013       1  0 22:17 ?        00:00:00 /lib/systemd/systemd --user
garlick     3014    3013  0 22:17 ?        00:00:00 (sd-pam)
garlick     3031    3010  0 22:17 ?        00:00:00 sshd: garlick@pts/0
garlick     3032    3031  0 22:17 pts/0    00:00:00 -bash
garlick     3054       1 99 22:17 ?        00:00:20 stress --cpu 4
garlick     3055       1 99 22:17 ?        00:00:20 stress --cpu 4
garlick     3056       1 99 22:17 ?        00:00:20 stress --cpu 4
garlick     3057       1 99 22:17 ?        00:00:20 stress --cpu 4
garlick     3064    3032  0 22:17 pts/0    00:00:00 ps -fu garlick
```
job log says
```
 garlick@picl0:~$ flux job eventlog ƒedFEJvZGb
1654319858.678003 submit userid=5588 urgency=16 flags=0
1654319858.693491 depend
1654319858.693580 priority priority=16
1654319858.700633 alloc annotations={"sched":{"resource_summary":"rank2/core[0-3]"}}
1654319858.701058 prolog-start description="job-manager.prolog"
1654319859.109224 prolog-finish description="job-manager.prolog" status=0
1654319859.121955 start
1654319869.136784 exception type="timeout" severity=0 userid=-1 note="resource allocation expired"
1654319874.174598 finish status=9
1654319874.175427 epilog-start description="job-manager.epilog"
1654319874.216140 release ranks="all" final=true
1654319874.680700 epilog-finish description="job-manager.epilog" status=0
1654319874.682091 free
1654319874.682163 clean
```

grondo:
Can you try with `SIGINT`? I can't reproduce this on my system.

What could be happening is that your job tasks are ignoring `SIGALRM`. When the shell receives a signal it sends it to the tasks it is managing. Of course, the shell can't catch `SIGKILL`, so only the shell gets killed when this signal is sent (after `SIGALRM` fails).

I think the IMP should probably signal the shell's process group for `SIGKILL` at least. Long term this can be handled a bit more cleanly since system instance jobs will be in their own cgroup. I'll open a separate issue since this is not related to the issue with unified cgroups.

garlick:
Works! 
```
garlick@picl0:~$ flux mini run -N1 sleep 3600
^Cflux-job: one more ctrl-C within 2s to cancel or ctrl-Z to detach
^Cflux-job: one more ctrl-C within 2s to cancel or ctrl-Z to detach
^C40.320s: job.exception type=cancel severity=0 interrupted by ctrl-C
flux-job: task(s) exited with exit code 143

garlick@picl0:~$ flux dmesg
[snip]
2022-06-04T15:00:07.615959Z job-exec.debug[0]: exec aborted: id=5404032558432256
2022-06-04T15:00:07.616134Z job-exec.debug[0]: exec_kill: 5404032558432256: signal 15
```
(and I verified that both shell and sleep were cleaned up)

garlick:
Oh sorry you said SIGINT.  OK this works too:
```
garlick@picl0:~$ flux mini submit -N1 sleep 3600
ƒjEbfMMH3m
garlick@picl0:~$ flux job kill --signal=SIGINT ƒjEbfMMH3m
```

grondo:
Oh, actually I meant either. Thanks!

So I guess the remaining question is if the `stress` program is ignoring SIGALRM. Does the timeout test work with `sleep` instead?

grondo:
Rebased.

garlick:
> So I guess the remaining question is if the stress program is ignoring SIGALRM. Does the timeout test work with sleep instead?

Confirmed, the timeout test  works with sleep.

It looks like `stress` does not block SIGLARM but it does fork. 
- first `flux job kill -14 ID` terminates the parent `stress`;  child reparents to init, shell continues running
- second `flux job kill -14 ID` does nothing
- manual `kill -14` of child `stress` terminates it; the shell terminates and job ends

grondo:
Ah, I had thought the shell used `killpg()` to deliver an intercepted signal to job tasks, but I guess this has never been the case.

We could try this change in the shell, though I'm not sure what other things might break:
(There is, for example, https://github.com/flux-framework/flux-core/issues/3445)

```diff
diff --git a/src/shell/task.c b/src/shell/task.c
index 19ca0f215..42f7e737c 100644
--- a/src/shell/task.c
+++ b/src/shell/task.c
@@ -193,7 +193,7 @@ int shell_task_start (struct shell_task *task,
                       shell_task_completion_f cb,
                       void *arg)
 {
-    int flags = 0;
+    int flags = FLUX_SUBPROCESS_FLAGS_SETPGRP;
     flux_subprocess_hooks_t hooks = {
         .pre_exec = subproc_preexec_hook,
         .pre_exec_arg = task,
```

grondo:
Well, that doesn't work since `setpgrp(2)` fails when a pty is active since the process is already a process group leader due to the use of `setsid(2)`. I think we'll need this instead:

```diff
diff --git a/src/common/libsubprocess/local.c b/src/common/libsubprocess/local.c
index 20bece963..d6d682aab 100644
--- a/src/common/libsubprocess/local.c
+++ b/src/common/libsubprocess/local.c
@@ -514,7 +514,8 @@ static int local_child (flux_subprocess_t *p)
         p->in_hook = false;
     }
 
-    if (p->flags & FLUX_SUBPROCESS_FLAGS_SETPGRP) {
+    if (p->flags & FLUX_SUBPROCESS_FLAGS_SETPGRP
+        && getpgrp () != getpid ()) {
         if (setpgrp () < 0) {
             fprintf (stderr, "setpgrp: %s\n", strerror (errno));
             _exit (1);
diff --git a/src/shell/task.c b/src/shell/task.c
index 19ca0f215..42f7e737c 100644
--- a/src/shell/task.c
+++ b/src/shell/task.c
@@ -193,7 +193,7 @@ int shell_task_start (struct shell_task *task,
                       shell_task_completion_f cb,
                       void *arg)
 {
-    int flags = 0;
+    int flags = FLUX_SUBPROCESS_FLAGS_SETPGRP;
     flux_subprocess_hooks_t hooks = {
         .pre_exec = subproc_preexec_hook,
         .pre_exec_arg = task,
```

I tried this change with the `stress` reproducer and both canceling the job as well as the timeout (`SIGALRM`) case worked (both the parent and forked `stress` child were terminated).

I'll open an issue over in flux-core.

garlick (approved):
LGTM!

I did have one question - if other parts of flux-security, for example a containment plugin, might need to differentiate cgroups versions, would it make sense to have a configure time check instead of the runtime check in the kill function?

Feel free to ignore - this solves the problem at hand!

grondo:
>  if other parts of flux-security, for example a containment plugin, might need to differentiate cgroups versions, would it make sense to have a configure time check instead of the runtime check in the kill function?

I don't think a configure time check would work since unified vs legacy cgroups is selected via a kernel command line option on systems where it is optional (e.g. RHEL 8 and its equivalents). However, we could move the check to a libutil function once it is needed.

PR #143, opened 2022-06-04
fix readthedocs build

grondo:
This is an attempt to fix the readthedocs build. Posting a PR so the build actually occurs.


garlick (approved):
LGTM

PR #150, opened 2022-06-30
allow privileged IMP to linger during a job

### LINKED ISSUE #147: imp: consider allowing privileged IMP to remain for duration of job
Currently, when `flux-imp exec` is used to start a multi-user job, the IMP execs the configured job shell directly after establishing new user credentials, and therefore no privileged process is running for the duration of the job. This was a design choice motivated by limiting the existence of privileged code to an absolute minimum.

However, the lack of a place to perform privileged operations at the end of a job is turning out to be a problem when we started looking into enabling support for more advanced IMP exec features, such as user session management via PAM. An application that wants to use PAM for session management is supposed to call [`pam_close_session(3)`](https://linux.die.net/man/3/pam_close_session) after the user session terminates if it called [`pam_open_session(3)`](https://linux.die.net/man/3/pam_open_session) to establish the session. While it is unlikely that any of the PAM modules that would typically be used for Flux jobs would have any session close operations (`pam_limits`, `pam_loginuid` for example), it is an unknown if executing directly from a PAM application after a session has been opened will cause any issues. Also, it would be inevitable that sites may configure a PAM session module for Flux which does require session close operations, and it is unlikely that we could provide a decent error in this case -- things would just not work, or worst case, a vulnerability would be introduced.

Even if PAM session management were not an issue, we would likely run into the requirement of some privileged cleanup or teardown operation once we started implementing IMP specific plugin support, so we would likely run into this deficiency soon enough.

A proposed solution here would be to have the privileged portion of `flux-imp` remain for the duration of the job. This  would require the addition of a `fork()` and `wait()` in the IMP's privileged process, and the child would then establish user credentials and exec. The IMP would then become the parent of the job shell, and would wait until the job shell completes, call any finalizing routines such as `pam_close_session(3)` and exit with the exit code of the job shell.

Some other random thoughts to keep in mind:
 - The IMP will now be signaled by `flux-imp kill` instead of the job shell, the IMP should forward any signals to its own process group, or `flux-imp kill PID` should be modified to send a signal to the child process of any `flux-imp`.
 - The IMP should take care to close all file descriptors which were handed off to the job shell, or otherwise ensure that no untrusted data could be established in its address space via this route.

grondo:
This is a WIP until some sharness tests are added.

This PR changes the behavior of the IMP to linger during a job. That is, instead of the privileged IMP process changing user credentials and executing the job shell during `flux-imp exec`, the process now forks a copy of itself and _that_ process does the user transition and exec of the job shell. The IMP then waits for the child to exit and itself exits with a modified wait status of the child.

This gives us a place to put any privileged deconstructors or finalization routines, e.g. we now have a spot to call `pam_close_session(3)` if the IMP were to use `pam_open_session(3)` before invoking the user's job.

Multiuser jobs will now have the following process tree form (e.g. for a 2 task sleep job)
```
flux-imp(root)───flux-shell(grondo)───2*[sleep]
```
Since the `flux-imp` now has the PID that libsubprocess or libsdprocess associates with the job, signals may be delivered to it instead of the flux-shell. Since the intent is for signals to be delivered to the shell and in turn the processes of the job, there are two other requisite changes to the IMP in this PR:

 * The IMP lingerer gets a signal handler for common signals that simply forward the signal to the child
 * `flux-imp kill PID` now additionally checks if `PID` is a `flux-imp` process owned by `root`, and redirects signals to that process' children if so.

Fixes #147


grondo:
I've removed WIP from this PR and it is ready for a full review.

I added a test in `t2001-imp-kill.t` to test that `flux-imp kill SIGNAL PID` avoids killing another `flux-imp` process running as root, but I can't find a sane way to run this test during CI. To be functional, it requires that a setuid IMP be run inside a cgroup that is owned by the calling user (i.e. the userid running the test). This isn't the case in all our docker builds since the docker container itself runs inside a cgroup owned by root as it is started by the docker daemon. Locally, I can run and verify the test using `systemd-run --user --pipe ./t2001-imp-kill.t`, but this also isn't possible inside CI since systemd is not running.

For now, maybe we can deal with the skipped test, though I encourage reviewers to ensure it works by executing it via `systemd-run`, or it may work on a desktop where you are logged in via logind, which will run all user processes under the user systemd instance.

src/imp/pidinfo.c
```diff
@@ -156,6 +194,64 @@ static uid_t pid_owner (pid_t pid)
...
+    char path [128];
+    char *line = NULL;
+    const int len = sizeof (path);
+    size_t size = 0;
+    FILE *fp = NULL;
+    pid_t ppid = -1;
+
+    int n = snprintf (path, len, "/proc/%ju/status", (uintmax_t) pid);
+    if ((n < 0) || (n >= len) || !(fp = fopen (path, "r")))
+        return (pid_t) -1;
```
garlick, line 237: doesn't set errno on buffer overflow.

src/imp/pidinfo.c
```diff
@@ -156,6 +194,64 @@ static uid_t pid_owner (pid_t pid)
...
+            while (isspace (*p))
+                ++p;
+            if (parse_pid (p, &ppid) < 0)
+                goto out;
+            break;
+        }
+    }
+out:
+    free (line);
+    fclose (fp);
```
garlick, line 251: clobbers errno

src/imp/pidinfo.c
```diff
@@ -156,6 +194,64 @@ static uid_t pid_owner (pid_t pid)
...
+    return ppid;
+}
+
 void pid_info_destroy (struct pid_info *pi)
 {
     free (pi);
```
garlick, line 257: clobbers errno

src/imp/pidinfo.c
```diff
@@ -180,12 +276,109 @@ struct pid_info *pid_info_create (pid_t pid)
...
+            rc = -1;
+            continue;
+        }
+        count++;
+    }
+    if (count == 0) {
+        imp_warn ("No children signaled for pid %d", (int) parent);
+        rc = -1;
+    }
+    closedir (dirp);
```
garlick, line 334: This function appears to need to set errno when it returns -1 (callers expect it?)  but
- continued loop iteration clobbers errno in the two "continue" blocks
- errno is not set if count == 0
- closedir() clobbers errno
- imp_warn() clobbers errno

src/imp/pidinfo.c
```diff
@@ -180,12 +276,109 @@ struct pid_info *pid_info_create (pid_t pid)
...
+    return rc;
+}
+
+int pid_kill_children (pid_t pid, int sig)
+{
```
garlick, line 339: Similar errno issues to pid_kill_children_fallback()

garlick (commented):
There are various errno issues in pidinfo.  I stopped reviewing after encountering those because I wanted to get clarification on the error handling strategy for this code.  If mostly it's just logging warnings to stderr, then maybe it's not so important to make sure errno is set by each function.  But the ones I called out appeared to be used with the assumption that errno is set on error.

src/imp/pidinfo.c
```diff
@@ -156,6 +194,64 @@ static uid_t pid_owner (pid_t pid)
...
+        digit = *s++ - '0';
+        val = val * 10 + digit;
+    } while (*s);
+
+    if (val < 0)
+        return -1;
+
+    *ppid = (pid_t) val;
+    return 0;
+}
```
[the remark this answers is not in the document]
grondo: I'm sure I had a good reason, though I can't remember it at the moment. Perhaps the tortuous error handling requirements of `strtoul`?

grondo:
> There are various errno issues in pidinfo. 

I think the fact that there was an error in this internal code is more important than the exact errno, which may or may not be helpful. These functions are not exported via any library. However, it does seem like this code which runs privileged warrants more careful handling, so I'll go through and attempt to fix these issues where it seems useful.

garlick:
Yeah, that's why I paused the review - it seemed unclear if setting errno was really all that helpful given the logging to stderr inline.

grondo:
Ok, I've made some significant changes here based on @garlick's feedback.

 * It was easier to kind of cleanup errno handling in `pidinfo.c` than make judgement calls about when it would be deemed necessary vs not. Since the original code in `pidinfo.c` is copied from `kill.c`, I added a commit that fixes some errno cloberring _after_ the move to avoid having to redo that part of the PR.
 * The "kill children" functions which are added as part of this PR now return the count of children successfully signaled processes instead of a more general 0 for success and -1 for failure. The IMP kill function that calls the function can then differentiate between no children signaled vs an error. `errno` should be valid now with any errors.
 * In support of the above the bullet, clean up errno handling in static functions. Do away with the snprintf() error checking since we know there is sufficient space in the destination buffer given that the only args format integers.
 * Expand testing to include tests of signaling processes with more than 1 child.

I went ahead and squashed the result - I'm not sure the fixup commits would have been helpful otherwise and in several cases I was manually amending commits anyway. There was enough change here that the review should be restarted.

Hopefully this is a little less sloppy, but I expect there may still be some corner cases identified in review.

src/imp/pidinfo.c
```diff
@@ -193,12 +301,117 @@ struct pid_info *pid_info_create (pid_t pid)
...
+                          (unsigned long) pid,
+                          strerror (errno));
+
+            }
+            continue;
+        }
+        if (ppid != parent)
+            continue;
+        if (kill (pid, sig) < 0) {
+            saved_errno = 0;
```
garlick, line 351: same here

src/imp/test/pidinfo.c
```diff
@@ -11,12 +11,140 @@
...
+    return pid;
+}
+
+static void pid_kill_tests (void)
+{
+    pid_t pid;
+    int status;
+
+    ok (pid_kill_children (-1, 0) < 0 && errno == EINVAL,
+        "pid_kill_children with invalid args returns EINVAL: got %d", errno);
```
garlick, line 90: errno needs to be cleared before tests that are verifying that it's been set, otherwise the result from the previous test can cause a false positive.

src/imp/kill.c
```diff
@@ -83,6 +83,15 @@ static void check_and_kill_process (struct imp_state *imp, pid_t pid, int sig)
...
                     strerror (errno));
+    if (sig == 0) {
+        printf ("pid=%ju owner=%ju cg_path=%s cg_owner=%ju",
+                 (uintmax_t) pid,
+                 (uintmax_t) p->pid_owner,
+                 p->cg_path,
+                 (uintmax_t) p->cg_owner);
+        pid_info_destroy (p);
+        return;
+    }
```
garlick, line 94: Hah, good idea!

garlick (approved):
LGTM!  Just a couple lingering errno issues.

src/imp/pidinfo.c
```diff
@@ -193,12 +301,117 @@ struct pid_info *pid_info_create (pid_t pid)
...
+        if (parse_pid (dent->d_name, &pid) < 0)
+            continue;
+        if ((ppid = pid_ppid (pid)) < 0) {
+            /* ENOENT is an expected error since a process on the system
+             *  could have exited between when we read the /proc dirents
+             *  and when we are checking for /proc/PID/status.
+             */
+            if (errno != ENOENT) {
+                saved_errno = 0;
+                rc = -1;
```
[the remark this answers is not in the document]
grondo: Yes! Good catch, thanks.

grondo:
Thanks @garlick! I've squashed and pushed fixes for the remaining issues you called out and will set MWP.

dun (commented):
Looks good! Just a few minor nitpicks. :+1:

src/imp/pidinfo.c
```diff
@@ -0,0 +1,417 @@
...
+        fprintf (stderr, "fopen %s: %s\n", path, strerror (errno));
+        return (pid_t) -1;
+    }
+
+    while ((n = getline (&line, &size, fp)) >= 0) {
+        if (strncmp (line, "PPid:", 5) == 0) {
+            char *p = line + 5;
+            while (isspace (*p))
+                ++p;
+            if (parse_pid (p, &ppid) < 0)
```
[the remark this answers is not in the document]
grondo: Oh, yeah. Looks like perhaps there was some other code here during the times of yore. I'll fix.

Issue #148, opened 2022-06-27
imp: add PAM session management support to IMP exec subcommand


grondo:
In a Flux system instance, the Flux IMP is the piece of software which allows the unprivileged system instance owner (typically the `flux` user) to transition user credentials to the _guest_ users that are running jobs. Since the IMP is the only component of a system instance which is run with elevated privileges, any job or user setup which may require these privileges must take place in the IMP before it transitions userids and calls `exec(2)`.

One useful way to allow runtime configuration of IMP behavior would be to turn the IMP into a [Linux PAM](https://en.wikipedia.org/wiki/Linux_PAM) enabled application. This would entail calling into the [PAM API](https://www.man7.org/linux/man-pages/man3/pam.3.html) while the IMP is still running setuid root. I believe it would be sufficient to just execute the session management API calls, since the IMP has already authenticated the user via the signed J (therefore no authentication phase needed), and it is presumed that user's account is valid (no account management phase required). This would involve calling `pam_open_session(3)` and `pam_close_session(3)` along with necessary other PAM API setup and `pam_set_item(3)` calls etc.

Note; In order for the IMP to be able to call `pam_close_session(3)`, we'll need a solution for #147.

garlick:
Resolved by #151 I think.  Open new issues if there's more to do.

PR #151, opened 2022-08-05
imp: exec: add PAM session support

wihobbs:
This is being put up as WIP for comments and review. It should not be merged until after #150.

This provides a solution to flux-framework/flux-pmix#40 and #148 where users have hit issues running on system-instance clusters where max locked memory must be increased. A few things to note here:
- The system instance will now require a PAM configuration file in `/etc/pam.d/flux` and (presumably) a `limits.conf` file with hard and soft resource limits. This should be added to the Flux Administrator's Guide.
- When running in a container, the `pam_limits.so` dependency will fail if a `limits.conf` file does not exist, a `/etc/pam.d/flux` file does not exist, or if the configured limits exceed the limits allocated to the container. This may require changes to the fluxorama containers to establish PAM support.

Potential to-do items:
- Slurm's PAM session configuration includes a header definition check `HAVE_PAM` which skips the PAM start and finish operations if these calls aren't supported by the system.
- The session configuration currently sets credentials (in my container) for user `flux`. The limits are the same, and the tests seem to work, but I don't know if this is what we want or not.

### DIFF
--- a/src/imp/exec/exec.c
+++ b/src/imp/exec/exec.c
@@ -33,7 +33,6 @@
 #include <wait.h>
 #include <jansson.h>
 
-
 #include "src/libutil/kv.h"
 #include "src/lib/context.h"
 #include "src/lib/sign.h"
@@ -44,6 +43,10 @@
 #include "privsep.h"
 #include "passwd.h"
 #include "user.h"
+
+#if HAVE_PAM
+#include "pam.h"
+#endif
 
 struct imp_exec {
     struct passwd *imp_pwd;
@@ -94,6 +97,15 @@ static bool imp_exec_unprivileged_allowed (struct imp_exec *exec)
 {
     return cf_bool (cf_get_in (exec->conf, "allow-unprivileged-exec"));
 }
+
+
+#if HAVE_PAM
+/* Check for PAM support, but default to not using PAM for now.
+ */
+static bool imp_supports_pam (struct imp_exec *exec) {
+    return cf_bool (cf_get_in (exec->conf, "pam-support"));
+}
+#endif
 
 static void imp_exec_destroy (struct imp_exec *exec)
 {
@@ -289,7 +301,14 @@ int imp_exec_privileged (struct imp_state *imp, struct kv *kv)
     if (privsep_wait (imp->ps) < 0)
         exit (1);
 
+#if HAVE_PAM
     /* Call privileged IMP plugins/containment */
+    if (imp_supports_pam (exec)) {
+        struct passwd *user_pwd = passwd_from_uid (exec->userid);
+        if (pam_setup (user_pwd->pw_name) < 0)
+            imp_die (1, "exec: PAM stack failure");
+    }
+#endif /* HAVE_PAM */
 
     /* Block signals so parent IMP isn't unduly terminated */
     sigblock_all ();
@@ -320,7 +339,11 @@ int imp_exec_privileged (struct imp_state *imp, struct kv *kv)
             imp_die (1, "waitpid: %s", strerror (errno));
     }
 
+#if HAVE_PAM
     /* Call privliged IMP plugins/containment finalization */
+    if (imp_supports_pam (exec))
+        pam_finish ();
+#endif /* HAVE_PAM */
 
     /* Exit with status of the child process */
     if (WIFEXITED (status))
--- a/src/imp/exec/pam.c
+++ b/src/imp/exec/pam.c
@@ -1,0 +1,83 @@
+/************************************************************\
+ * Copyright 2022 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+#include "pam.h"
+
+#include <stddef.h>
+#include <stdio.h>
+
+#include "imp_log.h"
+
+#include <security/pam_appl.h>
+#include <security/pam_misc.h>
+
+static pam_handle_t *pam_h = NULL;
+
+int pam_setup (const char *user)
+{
+    struct pam_conv conv = {misc_conv, NULL};
+    int rc;
+
+    if ((rc = pam_start ("flux",
+                         user,
+                         &conv,
+                         &pam_h)) != PAM_SUCCESS) {
+        imp_warn ("pam_start: %s", pam_strerror (NULL, rc));
+        goto fail1;
+    }
+    if ((rc = pam_set_item (pam_h, PAM_USER, user)) != PAM_SUCCESS) {
+        imp_warn ("pam_set_item USER: %s", pam_strerror (pam_h, rc));
+        goto fail2;
+    }
+    if ((rc = pam_set_item (pam_h, PAM_RUSER, user)) != PAM_SUCCESS) {
+        imp_warn ("pam_set_item RUSER: %s", pam_strerror (pam_h, rc));
+        goto fail2;
+    }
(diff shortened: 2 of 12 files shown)


src/imp/exec/pam.h
```diff
@@ -0,0 +1,27 @@
...
+\************************************************************/
+
+#ifndef HAVE_IMP_PAM_H_
+#define HAVE_IMP_PAM_H_ 1
+
+#include "imp_log.h"
+#include <security/pam_appl.h>
+#include <security/pam_misc.h>
+#include <stddef.h>
+#include <stdio.h>
```
grondo, line 18: All of these includes can be moved into `pam.c`

src/imp/exec/pam.h
```diff
@@ -0,0 +1,27 @@
...
+
+#define FLUX_SERVICE_PAM "flux"
```
grondo, line 20: This define should also be moved into `pam.c` I think, since it should be private to that file.

src/imp/exec/pam.c
```diff
@@ -0,0 +1,67 @@
...
+
+static pam_handle_t *pam_h = NULL;
+
+
+int pam_setup (char *user)
+{
+    struct pam_conv conv = {misc_conv, NULL};
+    int rc = 0;
+    if ((rc = pam_start (FLUX_SERVICE_PAM, user, &conv, &pam_h)) != PAM_SUCCESS) {
+        printf ("pam_start: %s", pam_strerror (NULL, rc));
```
grondo, line 21: Missed one `printf` :-)

src/imp/exec/pam.c
```diff
@@ -0,0 +1,67 @@
...
+        goto fail1;
+    } else if ((rc = pam_set_item (pam_h, PAM_USER, user)) != PAM_SUCCESS) {
```
grondo, line 23: I know this style is probably cribbed from Slurm, but since there's a `goto fail1;`, the `else if` is unnecessary here. Let's remove the `else if` and just have
```suggestion
        goto fail1;
    }
    if ((rc = pam_set_item (pam_h, PAM_USER, user)) != PAM_SUCCESS) {
```

grondo (commented):
Awesome!

Before merging I think we'll also want to have an IMP config parameter to enable PAM, since as you found, the PAM setup will fail if there is no `/etc/pam.d/flux`. We'll also need a way as you surmised to compile dummy calls if the IMP is not compiled with PAM support.

src/imp/exec/exec.c
```diff
@@ -288,6 +288,7 @@ int imp_exec_privileged (struct imp_state *imp, struct kv *kv)
         exit (1);
 
     /* Call privileged IMP plugins/containment */
+    pam_setup ((char *)((exec->imp_pwd)->pw_name));
```
grondo, line 291: Oh, also we should probably check the return code here.

grondo:
> The session configuration currently sets credentials (in my container) for user flux. The limits are the same, and the tests seem to work, but I don't know if this is what we want or not.

Ah, I only just caught this. Yeah, we'll need to run with the job's user, not the user running the IMP (user `flux`).
This works with `pam_limits` because all users on the system share the same `limits.conf` in our test scenario, however, calling into PAM with the `flux` user instead of the job's user could cause user jobs to inherit permissions of the `flux` user.

To get the username of the job user, we can initialize a `struct passwd` with
```c
    struct passwd *user_pwd = passwd_from_uid (exec->userid);

    /* Then, later... */
    pam_setup (user_pwd->pw_name);
```

Maybe for clarity, we should add an `exec->user_pwd` in addition to `exec->imp_pwd`.

src/imp/exec/pam.c
wihobbs: If the `goto fail1` and `goto fail3` are only used once each, is it acceptable to just place the code for those statements where they're used?

grondo: Well, in this case that would cause some code duplication. Because in all cases you need to reset `pam_h` to `NULL`, all the `goto fail*` can reuse this same block. It is probably safer to keep the `goto` statements in this case, also, because in the future if another conditional is added, the future developer is less likely to forget some error cleanup on failure.

src/imp/exec/pam.c
```diff
@@ -0,0 +1,67 @@
...
+
+fail3:
+    pam_setcred (pam_h, PAM_DELETE_CRED);
+
+fail2:
+    pam_end (pam_h, rc);
+
+fail1:
+    pam_h = NULL;
+    imp_die (rc, "failed to initialize PAM stack");
```
grondo, line 46: I'd remove this `imp_die` statement here, return failure from this function, and have the IMP call `imp_die` after checking the `pam_setup()` return code.

grondo:
> t2000 is failing under sudo because we don't currently have a PAM stack configured in the docker containers. I'll look into this and see if it's a fix that can be easily added to this PR.

We don't want to _require_ a PAM config even if the IMP was built with PAM support, so a preferred fix here is to only call into the  PAM stack if a pam config  key is set in the IMP config TOML. Do you think you can figure out how to do that? If not I can help in a couple hours..

grondo:
The build is currently failing in CI because:
 * I don't think --enable-pam is passed on the configure cmdline
 * If --enable-pam is not specified, the PAM code is still built and thus fails to link since the PAM libraries are not in PAM_LIBS due to the first item

wihobbs:
Ah, ok, I understand what you mean. Let me look into this (toml file).

wihobbs:
I have a partial fix for this with some limitations/todo items:
- The IMP configuration file will now require a `pam-support = true` line to run with PAM support. `t2003` has been updated accordingly. The IMP, for now, will default to _not_ opening a PAM session.
- I'm still working on the global/docker configure update since, as you noted @grondo, without `#ifdef` guards in the PAM code, the entire build for flux-security will fail when run without `./configure --enable-pam`. 
- `make check` passes for me on bionic and el8 but not fedora. Looking into how to add the `pam-devel` package for this build.

grondo:
> make check passes for me on bionic and el8 but not fedora. Looking into how to add the pam-devel package for this build.

Don't worry about that, I'll get it since it requires changes in flux-core.

For now, we can compile with `--enable-pam` only in the images known to have PAM development packages.

wihobbs:
@grondo I separated out the testsuite fixup from the PAM session commit for a cleaner commit history.

grondo:
I've reworked this PR a bit, added more tests, and added documentation for the  `pam-support` option. I've squashed some very minimal (style, support for compiling without PAM, and minor other updates) to @wihobbs'  PAM support commit 8faef9c.

This is probably now ready for an outside review.

garlick (approved):
This LGTM!

I did set this up with some altered `security.conf` limits on my test system and it worked as expected.

grondo:
Excellent! Thanks for the testing. Setting MWP.

src/imp/exec/pam.c
```diff
@@ -0,0 +1,83 @@
...
+        }
+        if ((rc = pam_end (pam_h, rc)) != PAM_SUCCESS) {
+            imp_warn ("pam_end: %s", pam_strerror (pam_h, rc));
+        }
+        pam_h = NULL;
+    }
+}
+
+/*
+ *  vi: sw=4 ts=4 expandtab
```
wihobbs, line 82: Just for my own training -- what is this note doing? Is it something with `vi` configurations for space and tab length? If so, why is it needed here? (I've just never seen this anywhere before.)

garlick: Yup that's what it is.  See also: https://vim.fandom.com/wiki/Modeline_magic

We've been using that in Flux source files from the beginning.  Although it's not helpful to contributors who don't use vi, several of the major contributors to flux do use vi and it helps keep us in line and save keystrokes.

wihobbs: Thanks @garlick!

PR #161, opened 2023-04-15
build: add 'make deb' target to build test debian package

grondo:
This PR adds debian packaging metadata files under a `debian` directory along with a script in `scripts/debbuild.sh` which creates a test deb package from the current build directory.

A `make deb` Makefile rule is also added, which handles the small subtlety of executing the build script from a vpath build.

I didn't have the `deb` target depend on `dist` because there didn't seem to be much benefit (the `dist` target always runs, even if it is rebuilding the exact same tarball, and running it from `make` doesn't allow you to suppress the output). This also allows the script to be used standalone.


garlick:
Ooh this is going to be great.

I'm hitting a test failure in t2002-imp-run:
```
not ok 25 - flux-imp run will not run file with bad ownership
FAIL: t2002-imp-run.t 25 - flux-imp run will not run file with bad ownership
#	
#		$SUDO chown $USER $TESTDIR/test.sh &&
#		test_must_fail $SUDO $flux_imp run test &&
#		$SUDO chown root $TESTDIR/test.sh
#
```
Apparently this test presumes $USER is set to a login name, but `debbuild` sets USER to the full name.  So in my case, `sudo chown Jim` fails because my user name is `garlick`.  That seems like the test's fault, not the PR's.  This fixes it
```diff
diff --git a/t/t2002-imp-run.t b/t/t2002-imp-run.t
index 1cf18df..01e739b 100755
--- a/t/t2002-imp-run.t
+++ b/t/t2002-imp-run.t
@@ -13,6 +13,8 @@ test -n "$FLUX_TESTS_LOGFILE" && set -- "$@" --logfile
 flux_imp=${SHARNESS_BUILD_DIRECTORY}/src/imp/flux-imp
 sign=${SHARNESS_BUILD_DIRECTORY}/t/src/sign
 
+USER=$(id -un)
+
 echo "# Using ${flux_imp}"
 
 test_expect_success 'flux-imp run returns error when run with no args' '
```

grondo:
Nice catch, thanks! I can fix that when I get back to a computer, or feel free to push the fix to this branch.

garlick (approved):
Seems to work well!

grondo:
Thanks! I'll merge this and I'm working on a similar approach for flux-core.

grondo:
And thanks for fixing the test bug!

PR #162, opened 2023-04-16
deb: do not package config file

grondo:
Problem: An example sign.toml is packaged in the resulting package after 'make deb', but this may overwrite an existing admin-installed config.

Remove all config files from the test package created by 'make deb'.


grondo:
Thanks to @garlick for noticing this.

garlick (approved):
LGTM!

PR #163, opened 2023-04-21
WIP: imp: support helper command to get input to exec command

### LINKED ISSUE #105: imp: allow alternate method for getting "input" for exec command
In the currently proposed version of the imp `exec` command, the "input to the IMP" is expected as a JSON object on `stdin` (currently just *J*, but extensible to the other IMP input components).

This approach will not be very scalable, since the job-exec module has to write the input separately to each IMP subprocess (1 per shell). Since the input to the IMP is read by an unprivileged process, it seems safe to allow the calling instance to override how the IMP gets its input. E.g. one simple idea would be for the caller to supply a commandline which the IMP can execute to gets its input. Similarly, the caller could supply a DSO that the IMP loads and executes. In either case, the IMP could then be fetching J and possibly other input directly from the KVS, which may be more scalable.
(Of course, input on stdin would still be supported)

grondo:
This PR adds support for a new `FLUX_IMP_EXEC_HELPER` environment variable which, when present, allows the input to `flux-imp exec` to be obtained from a command provided by the instance using the IMP, instead of written to stdin.

Though it should be safe to use `popen(3)` to run the helper, since this occurs in the unprivileged IMP, the command in `FLUX_IMP_EXEC_HELPER` is split into tokens on whitespace and passed directly to `execvp()`  to avoid the shell interpreting the value resulting in unintended or malicious bugs.

Still a WIP until some testing on the flux-core side is complete.

Fixes #105.

### DIFF
--- a/src/imp/exec/exec.c
+++ b/src/imp/exec/exec.c
@@ -43,6 +43,7 @@
 #include "privsep.h"
 #include "passwd.h"
 #include "user.h"
+#include "safe_popen.h"
 
 #if HAVE_PAM
 #include "pam.h"
@@ -369,8 +370,31 @@ static void imp_exec_put_kv (struct imp_exec *exec,
         imp_die (1, "exec: Failed to set job shell arguments");
 }
 
+/*  Read IMP input using a helper process
+ */
+static void imp_exec_init_helper (struct imp_exec *exec,
+                                  char *helper)
+{
+        int status;
+        struct safe_popen *sp;
+
+        if (!(sp = safe_popen (helper)))
+            imp_die (1, "exec: failed to invoke helper: %s", helper);
+
+        imp_exec_init_stream (exec, safe_popen_fp (sp));
+
+        if (safe_popen_wait (sp, &status) < 0
+            || status != 0)
+            imp_die (1, "exec: helper %s failed with status=0x%04x",
+                     helper,
+                     status);
+
+        safe_popen_destroy (sp);
+}
+
 int imp_exec_unprivileged (struct imp_state *imp, struct kv *kv)
 {
+    char *helper;
     struct imp_exec *exec = imp_exec_create (imp);
     if (!exec)
         imp_die (1, "exec: initialization failure");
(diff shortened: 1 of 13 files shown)


src/imp/exec/exec.c
```diff
@@ -381,8 +407,15 @@ int imp_exec_unprivileged (struct imp_state *imp, struct kv *kv)
...
-    imp_exec_init_stream (exec, stdin);
+    if ((helper = getenv ("FLUX_IMP_EXEC_HELPER"))
+        && strlen (helper) > 0) {
+        /* Read input from helper command */
+        imp_exec_init_helper (exec, helper);
+    }
+    else {
+        /* Read input from stdin, cmdline: */
+        imp_exec_init_stream (exec, stdin);
+    }
```
garlick, line 418: Might be worth noting why an empty string value for FLUX_IMP_EXEC_HELPER is not treated as an error?

garlick (approved):
Nice - thank you!  This looks good to me.

I just had two very minor comments inline.

src/imp/exec/safe_popen.c
```diff
@@ -0,0 +1,127 @@
...
+        _exit (126);
+    }
+
+    /* Parent
+     */
+    if (!(sp->fp = fdopen (pfds[0], "r"))) {
+        imp_warn ("fdopen: %s", strerror (errno));
+        int saved_errno = errno;
+        (void) kill (sp->pid, SIGKILL);
+        errno = saved_errno;
```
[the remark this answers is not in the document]
grondo: Good suggestion!

src/imp/exec/exec.c
grondo: Actually that's a good question. Maybe it should just be an error? For some reason I was thinking it was idiomatic to have an empty environment variable be treated as unset, as it makes things like this possible `VARIABLE= command` to "unset" a possibly set `VARIABLE` for the execution of `command`. Maybe I'll just say that in the comment.

grondo: Upon deep reflection, probably it is most clear and safest to return an error if `FLUX_IMP_EXEC_HELPER` is empty. This wasn't used in testing so it is easy to make this change.

grondo:
Ok, force pushed the suggested fixes. An empty `FLUX_IMP_EXEC_HELPER` is now an error. Thanks! I'll set MWP.

PR #166, opened 2023-05-04
build: enable PAM in `make deb`

grondo:
The debian package created by `make deb` was accidentally missing `--enable-pam` so had no PAM support.

Furthermore, the IMP silently ignores the `pam-support = true` option in the config when built without PAM.

Add PAM development packages to the suggested requirements for this project and in the `debian/control` file, and add `--enable-pam` to the debian package build by default. Add an error if `pam-support=true` is specified but the IMP was not built with PAM enabled, and add a test for this error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,17 +5,19 @@ Independent project for Flux security code and APIs.
 
 flux-security requires the following packages to build:
 
-**redhat**	| **ubuntu**		| **version**
-----------	| ----------		| -----------
-autoconf	| autoconf		|
-automake	| automake		|
-libtool		| libtool		|
-make		| make			|
-pkgconfig	| pkg-config	|
-libsodium-devel	| libsodium-dev		| >= 1.0.14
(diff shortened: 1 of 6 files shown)


grondo:
readthedocs builds are failing again with
```
ImportError: urllib3 v2.0 only supports OpenSSL 1.1.1+, currently the 'ssl' module is compiled with OpenSSL 1.0.2n  7 Dec 2017. See: https://github.com/urllib3/urllib3/issues/2168
```

I'm not exactly sure how we work around this, seems to be a readthedocs issue. Maybe we can pin urllib<2 in doc/requirements.txt? Or maybe it will fix itself later today.

grondo:
For now I've pinned urllib3<2 to get past the build failures.

garlick (approved):
Looks good!  I tested the deb build on my system and the rlimit warnings went away.

PR #169, opened 2023-08-08
ci: update build matrix to more closely match flux-core

grondo:
The CI in flux-core is being modernized to drop el7, bionic, and fedora35 images, and add Debian bookworm and fedora38.

Make similar changes here so that we can ensure flux-security is tested on the same sets of software as in flux-core's CI.
Include some minor fixes caught by more modern compilers now used in CI.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -2,18 +2,6 @@ queue_rules:
   - name: default
     conditions:
       - base=master
-      - status-success="validate commits"
-      - status-success="bionic"
-      - status-success="bionic - 32 bit"
-      - status-success="bionic - gcc-8,distcheck"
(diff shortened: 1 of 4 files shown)


garlick (approved):
All looks good to me.  Sorry for some reason I missed that this was hanging out here.

grondo:
Bah, I forgot to update `.mergify.yml`. I'll just remove the status checks in there as we've done elsewhere. Sorry about that!

grondo:
Ok, fixed. This will need a manual merge now.

garlick (approved):
Making the merge button green again.

grondo:
Thanks!

PR #171, opened 2023-08-10
imp: fix segfault on missing user

garlick:
Here's a fix for the IMP segfault @ofaaland found, when a user is not in the password file.

I tested this on my test cluster.   Figuring out a CI test may require some time to get reacquainted with the flux-security test suite.

### DIFF
--- a/src/imp/exec/exec.c
+++ b/src/imp/exec/exec.c
@@ -57,7 +57,7 @@ struct imp_exec {
     flux_security_t *sec;
     const cf_t *conf;
 
-    uid_t userid;
+    struct passwd *user_pwd;
     json_t *input;
 
     const char *J;
@@ -113,6 +113,7 @@ static void imp_exec_destroy (struct imp_exec *exec)
     if (exec) {
         flux_security_destroy (exec->sec);
         json_decref (exec->input);
+        passwd_destroy (exec->user_pwd);
         passwd_destroy (exec->imp_pwd);
         kv_destroy (exec->args);
         free (exec);
@@ -123,7 +124,6 @@ static struct imp_exec *imp_exec_create (struct imp_state *imp)
 {
     struct imp_exec *exec = calloc (1, sizeof (*exec));
     if (exec) {
-        exec->userid = (uid_t) -1;
         exec->imp = imp;
         exec->sec = sec_init ();
         exec->conf = cf_get_in (imp->conf, "exec");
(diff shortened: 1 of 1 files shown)


garlick:
Oh what did github do with my stack trace!  It's all issue references now :-)

grondo:
Should the existence of the target userid be validated even if PAM is not enabled, i.e. earlier in the IMP exec setup?
It might be a security issue to switch the job to a userid that is locally invalid.

Also, we should probably get #169 reviewed and merged before this one to make sure we're testing with the same base images as flux-core.

garlick:
How's this fix look?  I moved the passwd lookup into `imp_exec_unwrap()` and for ease of memory management, made the `user_pwd` part of the `struct imp_exec`.

garlick:
oops, the code inside #if HAVE_PAM is wrong.  Hmm, let's see if CI catches it then I'll push a fix.

grondo:
Yeah, that is what I would have done as well.

garlick:
The fedora34 builder failed on this test.  Not sure if that could be related (I think not?):
```
2023-08-10T22:25:58.9676326Z expecting success: 
2023-08-10T22:25:58.9676587Z 	name=unpriv &&
2023-08-10T22:25:58.9676891Z 	cat <<-EOF >${name}.toml
2023-08-10T22:25:58.9677200Z 	allow-sudo = true
2023-08-10T22:25:58.9677445Z 	[exec]
2023-08-10T22:25:58.9677746Z 	allowed-users = [ "$(whoami)" ]
2023-08-10T22:25:58.9678007Z 	EOF
2023-08-10T22:25:58.9678242Z 	sleep 300 & pid=$! &&
2023-08-10T22:25:58.9678524Z 	FLUX_IMP_CONFIG_PATTERN=${name}.toml \
2023-08-10T22:25:58.9678872Z 	    $flux_imp kill 15 $pid >${name}.log 2>&1 &&
2023-08-10T22:25:58.9679183Z 	test_debug "cat ${name}.log" &&
2023-08-10T22:25:58.9679462Z 	test_expect_code 143 wait $pid
2023-08-10T22:25:58.9679646Z 
2023-08-10T22:25:58.9679830Z test_expect_code: command exited with 0, we wanted 143 wait 9373
2023-08-10T22:25:58.9680287Z not ok 7 - flux-imp kill: works in unpriv mode
```
But no compilation failed despite this fix needed, so I surmise we aren't building with pam in CI
```
diff --git a/src/imp/exec/exec.c b/src/imp/exec/exec.c
index 745b636..b746172 100644
--- a/src/imp/exec/exec.c
+++ b/src/imp/exec/exec.c
@@ -312,7 +312,7 @@ int imp_exec_privileged (struct imp_state *imp, struct kv *kv)
     /* Call privileged IMP plugins/containment */
     if (imp_supports_pam (exec)) {
 #if HAVE_PAM
-        if (pam_setup (user_pwd->pw_name) < 0)
+        if (pam_setup (exec->user_pwd->pw_name) < 0)
             imp_die (1, "exec: PAM stack failure");
 #else
         imp_die (1,
```

I'll go ahead and squash the fix and push it.

grondo:
I'll have to check we didn't drop pam in our test images!

garlick:
`configure --with-pam` is opt-in and I don't think we're configuring it that way anywhere.  Maybe it ought to be enabled if the pam library is found?

grondo:
I thought it was enabled in at least one CI builder, we even build in `pam_wrapper` for ci in 6f67da77a74a7950c4ca16ba32bc431ef6f96fbd. I guess we did that but didn't add `--with-pam` to any of the CI builds. lame.

garlick:
I dropped the extra test commit that we determined offline to be fruitless.  This is ready for a review.

grondo (approved):
Looks good! Thanks for this!

PR #172, opened 2023-08-11
ci: add `--enable-pam` to most CI builds

grondo:
This PR simply adds `--enable-pam` to many of the ci builders. We can still consider auto-enabling PAM in the build, but for now this is a simpler approach.


garlick (approved):
LGTM!

grondo:
Thanks!

PR #173, opened 2023-08-18
testsuite: cover rfc38 kv test vectors

garlick:
Problem: rfc38 specifies test vectors for kv, but these are not used by unit tests.
    
Add a new unit test section that covers these vectors.

Also, grow an internal static buffer in case very large doubles are ever encoded.

### DIFF
--- a/src/libutil/kv.c
+++ b/src/libutil/kv.c
@@ -204,7 +204,7 @@ static int kv_put_raw (struct kv *kv, const char *key, enum kv_type type,
 
 int kv_vput (struct kv *kv, const char *key, enum kv_type type, va_list ap)
 {
-    char s[80];
+    char s[512]; // -DBL_MAX with 6 digit precision + \0 is 318 characters
     const char *val = NULL;
 
     if (!kv || !valid_key (key))
--- a/src/libutil/kv.c
+++ b/src/libutil/kv.c
@@ -219,7 +219,7 @@ int kv_vput (struct kv *kv, const char *key, enum kv_type type, va_list ap)
             val = s;
             break;
         case KV_DOUBLE:
-            if (vsnprintf (s, sizeof (s), "%f", ap) >= sizeof (s))
+            if (vsnprintf (s, sizeof (s), "%.6f", ap) >= sizeof (s))
                 goto inval;
             val = s;
             break;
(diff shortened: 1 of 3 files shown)


src/libutil/kv.c
```diff
@@ -204,7 +204,7 @@ static int kv_put_raw (struct kv *kv, const char *key, enum kv_type type,
 
 int kv_vput (struct kv *kv, const char *key, enum kv_type type, va_list ap)
 {
-    char s[80];
+    char s[317]; // DBL_MAX with 6 digit precision + \0
```
grondo, line 207: Should one more byte be added to the buffer in case of `-DBL_MAX`? What is one more byte among 317 friends?

grondo (approved):
Nice improvement!

src/libutil/kv.c
garlick: I added -DBL_MAX to the test and some wiggle room to that buffer.

garlick:
I tacked on a commit to impose a 100MB limit on overall encoded length.  That seems like it might be good enough for any reasonable jobspec?

grondo:
> That seems like it might be good enough for any reasonable jobspec?

That sounds good to me!

I wonder what the IMP will emit when encountering this limit? Or will the limit actually be hit when trying to sign_wrap() a jobspec (that's better IMO)

grondo (approved):
This LGTM!

garlick:
Yes `sign_wrap()` should fail first.  However if an older library or hand built input were sent to the IMP, it would fail on unwrap.
Hmm, wondering if maybe the size limit ought to be done in a separate PR with tests added for wrap/unwrap too?

grondo:
>  wondering if maybe the size limit ought to be done in a separate PR with tests added for wrap/unwrap too?
Not a bad idea if you're up for it.

garlick:
OK pushed without the size limit stuff.

grondo (approved):
LGTM!

garlick:
Oh duh, when signing, kv is only used to encode the  HEADER portion of _J_, which consists of just the  _version_, _mechanism_, _userid_.

```
 * The output of flux_sign_wrap() is a NULL-terminated string of the
 * form HEADER.PAYLOAD.SIGNATURE, resembling JSC Compact Serialization,
 * described in Internet RFC 7515:
 *
 * HEADER
 *   Contains the userid and signing mechanism in an internal key-value
 *   format, base64 encoded.
 *
 * PAYLOAD
 *   User data, base64 encoded
 *
 * SIGNATURE
 *   Mechanism dependent signature over HEADER.PAYLOAD, a string that
 *   must not contain "." (delimiter).
```
Maybe another RFC is called for to describe this level.

Anyway the jobspec is not encoded in the HEADER.

PR #176, opened 2023-09-05
add release notes for v0.10.0

grondo:
We should probably create a new flux-security release to incorporate the fix for #171.

This just adds release notes for a flux-security v0.10.0 release. Note sure if we have more planned work for the next tag at this time.


garlick (approved):
Thanks!

grondo:
Thanks! I set MWP.

PR #182, opened 2023-11-16
port to musl and add alpine CI build

grondo:
This PR fixes some musl libc portability issues in flux-security and adds an alpine Linux CI build so we don't break it again.

### DIFF
--- a/src/libutil/test/cf.c
+++ b/src/libutil/test/cf.c
@@ -19,6 +19,7 @@
 #include <limits.h>
 #include <errno.h>
 #include <jansson.h>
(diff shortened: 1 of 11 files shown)


garlick (approved):
All looks good to me!

src/libutil/test/cf.c
```diff
@@ -79,7 +79,7 @@ static time_t strtotime (const char *s)
 {
```
garlick, line 80: Commit message typo here.  Says `%T-%m-%d`(twice) but should be `%Y-%m-%d`

PR #183, opened 2023-11-17
NEWS.md: add release notes for v0.11.0

grondo:
A tagged version of flux-security that builds on alpine will be required for a flux-core alpine build for CI, since the first thing that is done is to build flux-security in the image. Therefore, let's create a v0.11. 0 tag for his purpose. Not much changed besides the porting effort, but tags are cheap...

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,14 @@
+flux-security version 0.11.0 - 2023-12-17
+-----------------------------------------
+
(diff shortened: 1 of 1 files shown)


garlick (approved):
LGTM!

grondo:
Thanks! I'll set MWP and tag.

PR #186, opened 2024-10-30
imp: exec: send SIGKILL to cgroup/children on SIGUSR1

grondo:
This PR updates the IMP to translate SIGUSR1 to SIGKILL for `flux-imp exec` as defined in RFC 15.

Additionally, if possible, the IMP will signal all members of its cgroup (besides itself) when delivering SIGKILL.

Finally, to better handle returning the proper wait status to _its_ parent, when the child of the IMP is terminated by a signal, the IMP raises that same signal to itself.

This is WIP pending real system testing.

### DIFF
--- a/src/imp/cgroup.c
+++ b/src/imp/cgroup.c
@@ -34,6 +34,9 @@
 #ifndef CGROUP_SUPER_MAGIC
 #define CGROUP_SUPER_MAGIC 0x27e0eb
 #endif
+#ifndef CGROUP2_SUPER_MAGIC
+#define CGROUP2_SUPER_MAGIC 0x63677270
+#endif
 #include <signal.h>
 
 #include "src/libutil/strlcpy.h"
@@ -123,7 +126,6 @@ static int cgroup_init_mount_dir_and_type (struct cgroup_info *cg)
     if (statfs (cg->mount_dir, &fs) < 0)
         return -1;
 
-#ifdef CGROUP2_SUPER_MAGIC
     /* if cgroup2 fs mounted: unified hierarchy for all users of cgroupfs
      */
     if (fs.f_type == CGROUP2_SUPER_MAGIC)
@@ -141,7 +143,6 @@ static int cgroup_init_mount_dir_and_type (struct cgroup_info *cg)
     if (fs.f_type == CGROUP2_SUPER_MAGIC)
         return 0;
 
-#endif /* CGROUP2_SUPER_MAGIC */
     /*  O/w, if /sys/fs/cgroup is mounted as tmpfs, we need to check
      *   for /sys/fs/cgroup/systemd mounted as cgroupfs (legacy).
      */
--- a/src/imp/cgroup.c
+++ b/src/imp/cgroup.c
@@ -1,0 +1,193 @@
+/************************************************************\
+ * Copyright 2024 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif
+#include <stdio.h>
+#include <unistd.h>
+#include <stdlib.h>
+#include <stdint.h>
+#include <string.h>
+#include <errno.h>
+#include <dirent.h>
+#include <ctype.h>
+#include <stdbool.h>
+#include <limits.h>
+
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <sys/vfs.h>
+#ifdef HAVE_LINUX_MAGIC_H
+#include <linux/magic.h>
+#endif
+#ifndef TMPFS_MAGIC
+#define TMPFS_MAGIC 0x01021994 /* from linux/magic.h */
+#endif
+#ifndef CGROUP_SUPER_MAGIC
+#define CGROUP_SUPER_MAGIC 0x27e0eb
+#endif
+
+#include "src/libutil/strlcpy.h"
+
+#include "cgroup.h"
+#include "imp_log.h"
+
+/*
+ *  Look up the current cgroup relative path from /proc/self/cgroup.
+ *
+ *  If cgroup->unified is true, then look for the first entry where
+ *   'subsys' is an empty string.
+ *
+ *  Otherwise, use the `name=systemd` line.
+ *
+ *  See NOTES: /proc/[pid]/cgroup in cgroups(7).
+ */
+static int cgroup_init_path (struct cgroup_info *cgroup)
+{
+    int rc = -1;
+    int n;
+    FILE *fp;
+    size_t size = 0;
+    char *line = NULL;
+    int saved_errno;
+
+    if (!(fp = fopen ("/proc/self/cgroup", "r")))
+        return -1;
+
+    while ((n = getline (&line, &size, fp)) >= 0) {
+        char *nl;
+        char *relpath = NULL;
+        char *subsys = strchr (line, ':');
+        if ((nl = strchr (line, '\n')))
+            *nl = '\0';
+        if (subsys == NULL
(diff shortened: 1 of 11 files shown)


grondo:
Hm, readthedocs build seems to have been broken at some point as well. Will have to look into that.

src/imp/cgroup.c
```diff
@@ -0,0 +1,172 @@
...
+        *(relpath++) = '\0';
+        if (cgroup->unified || strcmp (subsys, "pids") == 0) {
+            int len = sizeof (cgroup->path);
+            n = snprintf (cgroup->path,
+                          len,
+                          "%s%s",
+                          cgroup->mount_dir,
+                          relpath);
+            if ((n > 0) && (n < len))
+                rc = 0;
```
garlick, line 85: I don't think n <= 0 is a possible failure.  Then if only checking for one condition, the n (re-)assignment could b e avoided.

src/imp/cgroup.c
```diff
@@ -0,0 +1,172 @@
...
+        return -1;
+
+    while ((n = getline (&line, &size, fp)) >= 0) {
+        char *nl;
+        char *relpath = NULL;
+        char *subsys = strchr (line, ':');
+        if ((nl = strchr (line, '\n')))
+            *nl = '\0';
+        if (subsys == NULL || *(++subsys) == '\0'
+            || !(relpath = strchr (subsys, ':')))
```
garlick, line 73: style: split to one conditional per line?

src/imp/cgroup.c
```diff
@@ -34,6 +34,7 @@
 #ifndef CGROUP_SUPER_MAGIC
```
garlick, line 34: commit message: s/delivery/deliver/

src/imp/cgroup.c
```diff
@@ -168,5 +169,41 @@ struct cgroup_info *cgroup_info_create (void)
     return cgroup;
 }
 
+int cgroup_kill (struct cgroup_info *cgroup, int sig)
+{
+    int count = 0;
+    int rc = 0;
+    int saved_errno = 0;
+    char path [PATH_MAX+15]; /* cgroup->path + "/cgroup.procs" */
```
garlick, line 177: Did you mean `PATH_MAX+1`?

src/imp/cgroup.c
```diff
@@ -168,5 +169,41 @@ struct cgroup_info *cgroup_info_create (void)
...
+    FILE *fp;
+    unsigned long child;
+    pid_t current_pid = getpid ();
+
+    (void) snprintf (path, sizeof (path), "%s/cgroup.procs", cgroup->path);
```
garlick, line 182: Check for overflow?

src/imp/signals.c
```diff
@@ -107,5 +108,16 @@ void imp_setup_signal_forwarding (struct imp_state *imp)
        imp_die (1, "failed to block signals: %s", strerror (errno));
 }
 
+void imp_raise (int sig)
+{
+    signal (sig, SIG_DFL);
+    if (raise (sig) == 0)
+        pause ();
```
garlick, line 115: Oh, nice!  I hadn't even thought about the fact that the wait status would not have WIFSIGNALED true etc. in this case.

src/test/docker/docker-run-checks.sh
```diff
@@ -177,7 +177,7 @@ else
     docker run --rm \
```
garlick, line 177: commit message: /expclitly/explicitly/

garlick (approved):
Great!

I have some comments but all superficial and/or optional.

src/imp/cgroup.c
grondo: Sorry, this code is actually copied from existing code. I'll clean that up.

grondo:
As part of this work I switched the cgroup controller that the IMP checks for from `name=systemd` to `pids`, since the latter can actually be used in testing. I had thought this was working, but after further investigation, that isn't quite right and the `pids` controller is unchanged for processes in a job managed by sdexec.

I think the correct thing (after perusing [cgroups(7)](https://man7.org/linux/man-pages/man7/cgroups.7.html)) is to use the `0::` line in `/proc/self/cgroup`, which I _think_ corresponds to the cgroups v2 hierarchy.

I'll have to modify this PR to make that change.

src/imp/cgroup.c
grondo: hm. cgroup->path is a buffer of size `PATH_MAX`. We're going to write 14 more characters so we need at least that much space or the compiler complains. Happy to do this another way.

src/imp/cgroup.c
grondo: Here we know we've added enough space to the buffer so it can't overflow. I guess `sprintf` could be used with a note in that case?

src/imp/cgroup.c
garlick: Oh right, I hate that.  Oh well.

grondo:
Ok, I think I've fixed the cgroup discovery here. The code is somewhat simplified, at startup the IMP:

 - checks if either `/sys/fs/cgroup` or `/sys/fs/cgroup/unified` are mounted as cgroup2
 - if not then check for a legacy (cgroup v1) mount at `/sys/fs/cgroup/systemd`
 - grab the current relative cgroup path for either case from `/proc/self/cgroup` (logic to find the correct one is now fixed)
 
I also had to add a workaround to remove leading `/..` from the relative cgroup path, which seems to pop up when run within a container (the relative path is relative to the container mount point on the host I think)

garlick:
I grabbed your snapshot from last night and put it on my test cluster.  This morning I made some minimal changes to job-exec on my `issue#6406` branch to send SIGUSR1 instead of imp-kill when shell termination methods fail.  Then I ran a test on my system by starting a long running job, sending the flux-shell a SIGSTOP, then canceling the job.  Looks good!  Flux on rank 0 logged:
```
[  +5.002092] job-exec[0]: Sending SIGKILL to job ƒ3L9MFPYEKZ
[ +10.005491] job-exec[0]: Sending SIGKILL to job ƒ3L9MFPYEKZ
[ +15.005766] job-exec[0]: Sending SIGKILL to job ƒ3L9MFPYEKZ
[ +20.005115] job-exec[0]: Sending SIGKILL to job ƒ3L9MFPYEKZ
[ +25.005382] job-exec[0]: Sending SIGKILL to job ƒ3L9MFPYEKZ
[ +25.005415] job-exec[0]: Sending SIGUSR1 to IMP for job ƒ3L9MFPYEKZ
[ +25.169307] job-manager[0]: housekeeping: ƒ3L9MFPYEKZ started
[ +25.201629] job-list[0]: purged 1 inactive jobs
[ +25.203373] job-manager[0]: purged 1 inactive jobs
[Nov01 08:43] job-manager[0]: housekeeping: ƒ3L9MFPYEKZ complete
```
On the compute node, systemd logged
```
Nov 01 08:41:05 picl1 systemd[796]: Starting imp-shell-1-f3L9MFPYEKZ.service - User workload...
Nov 01 08:41:05 picl1 systemd[796]: Started imp-shell-1-f3L9MFPYEKZ.service - User workload.
Nov 01 08:42:32 picl1 systemd[796]: imp-shell-1-f3L9MFPYEKZ.service: Sent signal SIGTERM to main process 907 (flux-imp) on client request.
Nov 01 08:42:57 picl1 systemd[796]: imp-shell-1-f3L9MFPYEKZ.service: Sent signal SIGUSR1 to main process 907 (flux-imp) on client request.
Nov 01 08:42:57 picl1 systemd[796]: imp-shell-1-f3L9MFPYEKZ.service: Main process exited, code=killed, status=9/KILL
Nov 01 08:42:57 picl1 systemd[796]: imp-shell-1-f3L9MFPYEKZ.service: Failed with result 'signal'.
```

grondo:
This morning I added one fix here: `cgroup_kill()` was always sending `SIGKILL` instead of the `sig` parameter. Fixed.

garlick (approved):
LGTM!

grondo:
Great, thanks! I'll set MWP. I'll post a PR to have the IMP wait until its cgroup is empty in a bit.

PR #187, opened 2024-11-01
imp: exec: wait for job cgroup to be empty before exiting

grondo:
This PR modifies `imp exec` to wait for an "empty" cgroup before exiting, where empty means `cgroup.procs` contains only the IMP's own PID.


garlick (approved):
LGTM!  Seems straightforward.

grondo:
Thanks I'll set MWP

PR #188, opened 2024-11-01
allow privileged IMP to linger during `flux-imp run` to support signal forwarding

grondo:
This PR changes the `flux-imp run` implementation to have the privileged IMP linger and delegate signals to its child, which the calling user would not otherwise have privilege to do.

WIP while I test this out locally.

### DIFF
--- a/src/imp/run.c
+++ b/src/imp/run.c
@@ -51,6 +51,7 @@
 
 #include <sys/types.h>
 #include <sys/stat.h>
+#include <sys/wait.h>
 #include <pwd.h>
 #include <signal.h>
 
@@ -61,6 +62,7 @@
 #include "imp_state.h"
 #include "impcmd.h"
 #include "privsep.h"
+#include "signals.h"
 
 extern char **environ;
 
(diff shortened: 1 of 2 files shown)


garlick (approved):
Well jeez, this looks straightforward also!  I'll approve assuming your testing pans out!

grondo:
Actually there's a couple issues still remaining here:
- test is broken because the `NO_CHAIN_LINT` prereq isn't set. When I fix that the test doesn't work :thinking: 
- the `imp run` implementation calls `setuid (geteuid ()); setgid (setegid ());` before calling the run target, which means the instance isn't actually able to deliver signals to the IMP process (so signal forwarding doesn't actually work)

grondo:
I had forgotten that `flux-imp run` set the real uid/gid to the effective uid/gid before executing the target command. This prevented signals from the calling uid from being delivered. I've moved the reset of the real uid/gid to the child process just before `execve(2)` and now signaling the `flux-imp run` process seems to be working.

I also figured out some issues with the test and that appears to now be working as well.

This could definitely use some more real world testing.

grondo:
Removed WIP now. This _seems_ to be working as designed now, and is much more robust than the `flux-imp kill` method. Probably would be good if @garlick verifies on his cluster before this gets merged though.

garlick:
Excellent!  Will try to get that done tomorrow morning.

garlick (approved):
I tried some basic sanity testing on my cluster, sending SIGTERM and SIGKILL to housekeeping with this, and everything worked!  LGTM!

grondo:
Thanks! I've set mwp

PR #193, opened 2024-11-04
configure: do not use distutils StrictVersion

grondo:
Problem: distutils StrictVersion is deprecated (along with distutils itself), but it is still used in testing for sphinx and docutils in configure.ac.

Drop distutils and use packaging.version.Version instead.


garlick (approved):
:+1:

grondo:
Thanks, I've set MWP. Sorry I missed this one!

PR #195, opened 2024-11-06
forward signals to entire process group in `flux-imp run`

### LINKED ISSUE #194: imp run doesn't forward (most) signals to all progeny
Problem: `flux run` forwards most signals to its direct child, and SIGUSR1 (SIGKILL) to the whole family, which is appropriate for the `flux shell` but less so for, say, a bourne shell script, where the bourne shell may be less diligent about cleaning up its children.

Change `flux run` signal forwarding behavior to try a little harder to clean up.

Note that this likely doesn't affect prolog, epilog, or housekeeping when they are configured to use systemd, since in that case, the shell script  executes `systemctl stop` in response to signals.  It does affect something like `flux exec --use-imp sh -c "sleep inf"` when the user types ^C to abort.

grondo:
This PR fixes #194 by allowing the IMP to forward signals to the entire process group with `flux-imp run`.

Additionally, an issue with blocked signals (specifically in `t2002-imp-run.t`) was fixed here  and a test for SIGINT handling added to the testsuite.

### DIFF
--- a/src/imp/run.c
+++ b/src/imp/run.c
@@ -199,6 +199,12 @@ imp_run (struct imp_state *imp,
         /* unblock all signals */
         imp_sigunblock_all ();
 
+        /* Place child in its own process group, so that parent IMP
+         * can signal the pgrp as a whole
+         */
+        if (setpgrp () < 0)
+            imp_die (1, "setpgrp: %s", strerror (errno));
+
         if (setuid (geteuid()) < 0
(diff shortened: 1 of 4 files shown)


garlick (approved):
LGTM!  and my testing (before the last force push) seems to be fine except for the extra log messages which were removed in the force push.

grondo:
Thanks! I set MWP.

PR #200, opened 2025-02-19
imp: use cgroup_kill when cgroup begins with flux-

garlick:
Problem: the cgroup name prefix of "imp-shell" is hard coded in the IMP to select a signal forwarding strategy that ought to apply to flux-housekeeping/epilog/prolog if we move them from the system to the user systemd instance.

Check for the "flux-" unit name prefix too.


garlick:
Thanks for the help on this one!  I'll set MWP.

garlick:
@mergifyio requeue

grondo:
Hm, somehow merge commits were disabled in this repo. I've re-enabled them, so let's try again:

@mergifyio refresh

grondo:
@mergifyio requeue

PR #203, opened 2025-02-27
imp run: wait for cgroup and call sd_notify()

garlick:
This prepares to let `imp run` run as the main process in a user systemd instance
- wait for cgroup to empty out before exiting
- call sd_notify() so Type=notify unit can be used (and allow NOTIFY_SOCKET to be passed thru the environment).

It is built on top of #201

### DIFF
--- a/src/imp/run.c
+++ b/src/imp/run.c
@@ -57,6 +57,7 @@
 
 #include "src/libutil/kv.h"
 #include "src/libutil/path.h"
+#include "src/libutil/sd_notify.h"
 
 #include "imp_log.h"
 #include "imp_state.h"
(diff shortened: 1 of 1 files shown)


src/imp/run.c
```diff
@@ -281,6 +281,7 @@ int imp_run_privileged (struct imp_state *imp,
     struct kv *kv_env;
     const char *name;
     const cf_t *cf_run;
+    const char *notify_socket = getenv ("NOTIFY_SOCKET");
```
grondo, line 284: Could instead add this variable to `run_env_var_allowed()` for consistency with `FLUX_JOB_ID` and `FLUX_JOB_USERID`.

If that isn't desirable, a comment here might help remind a reader why this variable is special (though admittedly at this point it's obvious)

grondo (dismissed):
LGTM! One suggestion inline.

src/imp/run.c
garlick: Hmm, i tried that originally and it didn't work for some reason and got myself a little confused in the process. Let me try again and see if I can run down what's happening.

garlick: Oh, the privileged imp process has its environment cleared, and the environment curated by the unprivileged imp process is passed through to `execve()`. But the imp itself, whose environment remains empty, will be making the `sd_notify()` calls so this does need to be handled differently.

I'll add a comment.

garlick:
Also need reapproval on this one :-)

grondo (approved):
Nice! Thank you!

PR #211, opened 2026-01-07
lib: avoid discarded-qualifiers error with gcc-16

### LINKED ISSUE #210: Build failure on Fedora Rawhide (GCC 16): const-correctness issues in sign.c
## Potential Problem

flux-security v0.14.0 fails to build on Fedora Rawhide (Fedora 44) potentially due to const-correctness warnings being treated as errors with GCC 16.

## Build Environment

- **Distribution**: Fedora Rawhide (44)
- **Compiler**: GCC 16.0.0-0.4.fc44
- **Build system**: COPR

## Real Error Messages

```
sign.c: In function ‘header_decode’:
sign.c:328:13: error: assignment discards ‘const’ qualifier from pointer target type [-Werror=discarded-qualifiers]
  328 |     if (!(p = strchr (input, '.'))) {
      |             ^
sign.c: In function ‘payload_decode_cpy’:
sign.c:368:13: error: assignment discards ‘const’ qualifier from pointer target type [-Werror=discarded-qualifiers]
  368 |     if (!(p = strchr (input, '.'))) {
      |             ^
cc1: all warnings being treated as errors
make[2]: *** [Makefile:846: sign.lo] Error 1
```

## Potential Root Cause

In C23 (and stricter GCC 16 enforcement), `strchr()` returns `const char*` when given a `const char*` input. In [`src/lib/sign.c`](https://github.com/flux-framework/flux-security/blob/master/src/lib/sign.c), the following pattern is used:

```c
static int header_decode(const char *input, ...)
{
    char *p;
    // ...
    if (!(p = strchr (input, '.'))) {  // input is const char*, strchr returns const char*
        // ...
    }
```

The assignment of `strchr()`'s return value (which is now `const char*` when the input is `const char*`) to a non-const `char *p` triggers `-Wdiscarded-qualifiers`, which becomes an error due to `-Werror`.

## Affected Code Locations

1. **`header_decode()`** - line ~328
2. **`payload_decode_cpy()`** - line ~368

## Potential Fix

Change the local variable declarations from `char *p` to `const char *p` in both functions:

```diff
 static int header_decode(const char *input, ...)
 {
-    char *p;
+    const char *p;
```

This should be safe since the pointer `p` is only used for reading/parsing, not for modification.

## Workaround

For packagers, adding `-Wno-error=discarded-qualifiers` to `CFLAGS` should allow the build to succeed:

```bash
export CFLAGS="${CFLAGS} -Wno-error=discarded-qualifiers"
```

## Full Build Log

The complete COPR build log is available at:
https://copr.fedorainfracloud.org/coprs/kushgupta/flux-security/build/9979806/

The spec file being used for `flux-security` & `flux-core` within these builds can be found below if you'd like to inspect for cruft/anomalies:
https://github.com/kush-gupt/flux-rpm

## Additional Context

This issue will likely affect all distributions that upgrade to GCC 16+ as they adopt stricter C23 const-correctness enforcement. 

I could also be **completely wrong here** since I had an AI help me debug this far, so feel free to take this issue with a grain of salt!

garlick:
Problem: on Fedora rawide (Fedora 44) with gcc-16, compilation fails with the following error:

sign.c: In function ‘header_decode’:
sign.c:328:13: error: assignment discards ‘const’ qualifier from pointer target type [-Werror=discarded-qualifiers]
  328 |     if (!(p = strchr (input, '.'))) {
      |             ^
sign.c: In function ‘payload_decode_cpy’:
sign.c:368:13: error: assignment discards ‘const’ qualifier from pointer target type [-Werror=discarded-qualifiers]
  368 |     if (!(p = strchr (input, '.'))) {
      |             ^
cc1: all warnings being treated as errors
make[2]: *** [Makefile:846: sign.lo] Error 1

Solution: declare some (char *) variables as (const char *).

Fixes #210

### DIFF
--- a/src/lib/sign.c
+++ b/src/lib/sign.c
@@ -315,9 +315,9 @@ const char *flux_sign_wrap (flux_security_t *ctx,
  * Return header on success or NULL on error with errno set.
  * Set 'endptr' to period ('.') delimiter following HEADER.
  */
-static struct kv *header_decode (const char *input, char **endptr)
+static struct kv *header_decode (const char *input, const char **endptr)
 {
-    char *p;
+    const char *p;
     const char *src;
     size_t srclen;
     void *dst;
@@ -358,9 +358,9 @@ error:
  * Return 0 on success, -1 on failure with errno set.
  */
 static int payload_decode_cpy (const char *input, void **buf, int *bufsz,
-                               char **endptr)
+                               const char **endptr)
 {
-    char *p;
+    const char *p;
     size_t dstlen;
     size_t srclen;
     const char *src;
@@ -412,7 +412,7 @@ static int sign_unwrap (flux_security_t *ctx,
     const char *mechanism;
     const struct sign_mech *mech;
     const cf_t *allowed_types;
-    char *endptr;
+    const char *endptr;
 
     if (!ctx || !input || !(flags == 0 || flags == FLUX_SIGN_NOVERIFY)) {
         errno = EINVAL;


garlick:
I tested this on fedora 40 docker image and then realized it was only gcc-14 and I need rawhide.  If convenient, would you mind giving this a try @kush-gupt?

kush-gupt:
Hey @garlick thanks for the quick fix! I tested under https://copr.fedorainfracloud.org/coprs/kushgupta/flux-security/build/9980014/ but there seems to be one missing change: in `payload_decode_cpy`, the local char *p on line 363 also needs to be `const char *p`. The function signature was updated but not the local variable. 

Could you try adding this? Throwing together a patch to see if this fully solves it on my end!

kush-gupt:
I got it green under https://copr.fedorainfracloud.org/coprs/kushgupta/flux-security/build/9980377/ with your patch & that 
`payload_decode_cpy` local var going from `char *p → const char *p`!

garlick:
Thanks!  I pushed that update.

garlick:
@chu11 would you mind giving this a review?

chu11 (approved):
LGTM!

PR #213, opened 2026-01-28
doc: remove reference to flux-mini(1)

wihobbs:
Problem: flux-mini(1) was lost to time but its references live on in the security documentation.

Update the documentation.

### DIFF
--- a/doc/man5/flux-config-security-sign.rst
+++ b/doc/man5/flux-config-security-sign.rst
@@ -6,7 +6,7 @@ flux-config-security-sign(5)
 DESCRIPTION
 ===========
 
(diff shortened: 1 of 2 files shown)


chu11 (dismissed):
good catches.  Only 1 minor suggestion, since it says "job submission tools" and "short lived", perhaps use "flux submit" as the example instead of "flux run".

chu11 (approved):
LGTM

wihobbs:
@chu11 Can you add the MWP label? I don't have that ability; this repo must not inherit maintainer privileges from the core team (IIRC pmix doesn't either).

PR #220, opened 2026-04-12
testsuite: fix cgroup mount dir detection in `t2000-imp-exec.t`

grondo:
The cgroup mount directory detection in `t2000-imp-exec.t` is flawed and could end up as empty on some distros. The test shoudl check for `cgroup2` in field 3 and not as the first field in `/proc/self/mounts`. 

This PR fixes the detection and adds a bail out if `CGROUP_MOUNT` is empty.

### DIFF
--- a/t/sharness.d/00-bail_out.sh
+++ b/t/sharness.d/00-bail_out.sh
@@ -1,0 +1,15 @@
+# Immediately bail out of the test script with the reason given in [$1].
+# See <https://testanything.org/tap-specification.html>.
+#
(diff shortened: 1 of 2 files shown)


garlick (approved):
Agreed, probably fragile.  LGTM!

p.s. that bail out function could be useful in flux-core!

grondo:
Thanks! I've set MWP. This was indeed causing testsuite failures on an Ubuntu 25.10 VM.

PR #222, opened 2026-04-14
fix minor issues discovered during security audit

grondo:
This PR fixes a series of minor issues discovered while using Claude Sonnet 4.6 to do a security-focused audit of the flux-security codebase.

### DIFF
--- a/src/imp/privsep.c
+++ b/src/imp/privsep.c
@@ -187,13 +187,12 @@ int privsep_wait (privsep_t *ps)
     int status = 0;
     if (privsep_is_parent (ps)) {
         if (ps->cpid > (pid_t) 0) {
-            kill (SIGTERM, ps->cpid);
             if (waitpid (ps->cpid, &status, 0) < 0)
                 status = -1;
             ps->cpid = (pid_t) -1;
         }
     }
-    return (status == 0 ? 0 : -1);
+    return ((WIFEXITED (status) && WEXITSTATUS (status) == 0) ? 0 : -1);
 }
 
 void privsep_destroy (privsep_t *ps)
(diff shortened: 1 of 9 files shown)


src/imp/privsep.c
```diff
@@ -187,13 +187,12 @@ int privsep_wait (privsep_t *ps)
...
         if (ps->cpid > (pid_t) 0) {
-            kill (SIGTERM, ps->cpid);
             if (waitpid (ps->cpid, &status, 0) < 0)
                 status = -1;
             ps->cpid = (pid_t) -1;
         }
     }
-    return (status == 0 ? 0 : -1);
+    return (WIFEXITED (status) && WEXITSTATUS (status) == 0 ? 0 : -1);
 }
```
garlick, line 203: I think you need parens around the && expression?

garlick (dismissed):
Excellent!  Just found one small item.

src/imp/privsep.c
grondo: Good suggestion, though I think && has precedence over `?:` so it will work as is. But parentheses will make that clearer!

grondo: Fixed.

grondo:
Addressed the comment and will set MWP. Thanks!

grondo:
For some reason github wants you to re-review @garlick 🤷

src/imp/privsep.c
garlick: Oh you're right, that was no problem (but yes, a little clearer for forgetful people)

garlick (approved):
Still great!

PR #223, opened 2026-04-16
MAINTAINERS.md: drop non-maintainers

garlick:
Problem: MAINTAINERS.md was copied from another repo with the assumption that the 'core' team had admin access, but that was not the case. Only organization owners had access here.

Cut back the MAINTAINERS.md list to Mark and Jim.


grondo (approved):
LGTM!

PR #227, opened 2026-04-21
doc: organize docs so some internals notes can be added

garlick:
I wanted to get the docs organized similar to flux-core with a "Resources for Flux Developers" section that could be amended with device containment info as that is developed.


grondo (approved):
LGTM!

garlick:
Thanks! Setting MWP.

PR #228, opened 2026-04-22
imp: add bpf device containment

### LINKED ISSUE #226: Support RFC 15 Device Containment
Problem: Device containment via systemd `DeviceAllow` and `DevicePolicy` can't be enforced by the `flux` user systemd instance because device containment is implemented in cgroupsv2 by loading bpf programs and attaching them to the cgroup, which requires `CAP_BPF` and `CAP_SYS_ADMIN`. There is therefore no support in systemd for delegating the handling of these properties to systemd user instances.
For details see discussion in  flux-framework/flux-core#7546.

The IMP exec helper (provided by flux-core) will fill in the [Device Containment](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_15.html#device-containment) section of the IMP's "resource owner options" input by reading `DeviceAllow` and `DevicePolicy` from the current cgroup. The IMP will need to read this section of its input, compose and load a suitable bpf program and attach it to the cgroup before launching the job shell.

garlick:
Problem: systemd running as the flux user cannot enforce `DeviceAllow` / `DevicePolicy` because it lacks appropriate privilege.  The IMP does have that privilege so we can implement it there.

See the  included design spec for design decisions made while building this.

I'm posting this early for quick feedback on the design doc.  The code at this point is first draft, completely untested.

Fixes #226


doc/guide/device-containment.rst
```diff
@@ -0,0 +1,199 @@
...
+currently ignored by systemd since the user instance lacks the
+privilege to enforce them.
+
+Per :doc:`RFC 15 <rfc:spec_15>` [7]_, the IMP takes its input from a helper
+program provided by flux-core, pointed to by :envvar:`FLUX_IMP_EXEC_HELPER`.
+The helper reads the systemd properties, interprets the policy, and produces
+a JSON object containing the signed job specification ``J`` and, when device
+containment is requested, a ``DeviceAllow`` list normalized to major/minor
+numeric form.  If ``DeviceAllow`` is present, access is restricted to the
+listed devices; if absent, all devices are accessible.  For example:
```
grondo: Might want to call out here that [baseline devices](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_15.html#baseline-devices) (i.e. those that would be included in the systemd `DevicePolicy=closed` policy) should be included in the `DeviceAllow` input to the IMP. I guess the imp helper will need to hard code these.

doc/guide/device-containment.rst
```diff
@@ -0,0 +1,199 @@
...
+
+.. code-block:: json
+
+    {
+      "J": "<signed jobspec>",
+      "DeviceAllow": [
+        { "type": "c", "major": 195, "minor": 0, "access": "rw" },
+        { "type": "c", "major": 195, "minor": -1, "access": "rw" }
+      ]
+    }
```
grondo, line 46: RFC 15 specifies that the only allowed keys in the input to the IMP are `J` and `options`:
https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_15.html#input-to-the-imp
and the [Device Containment](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_15.html#device-containment) specification needs an update to indicate the new `DeviceAllow` structure.

doc/guide/device-containment.rst
```diff
@@ -0,0 +1,199 @@
...
+form — is performed by the unprivileged helper before data crosses the
+privsep boundary.  The privileged IMP parent receives a pre-validated
+allowlist in a simple numeric form, minimizing the complexity and attack
+surface of the code that runs with elevated privilege.
+
+**Macro parameter naming.**  The BPF instruction-building macros avoid
+parameter names ``off`` and ``imm`` because the C preprocessor
+substitutes macro parameters before the compiler sees struct field
+names.  A parameter named ``off`` would be expanded inside ``.off =
+(off)`` designators, producing invalid syntax.  The parameters are
```
grondo: This one needs more explanation I think. What are `poff` and `pimm`? Is this detail really necessary in this doc?

doc/guide/device-containment.rst
```diff
@@ -0,0 +1,199 @@
...
+        r0 = 0; exit  (default deny)
+
+The ``access_type`` field of :c:struct:`bpf_cgroup_dev_ctx` packs the
+device type into the lower 16 bits (:c:macro:`BPF_DEVCG_DEV_BLOCK`,
+:c:macro:`BPF_DEVCG_DEV_CHAR`) and the requested access into the upper
+16 bits (:c:macro:`BPF_DEVCG_ACC_MKNOD`, :c:macro:`BPF_DEVCG_ACC_READ`,
+:c:macro:`BPF_DEVCG_ACC_WRITE`).
+
+Each allow entry emits 10 instructions without a minor check, or 11
+with one; the jump offsets within each entry are computed accordingly.
```
grondo, line 142: If there's a reason the number of instructions needs to be explicit it would be interesting to know that here.

grondo (commented):
Wow! It is really nice to have some clear design docs up front!
I've made a few comments inline.

doc/guide/device-containment.rst
garlick: Naa, useless detail, will drop.

doc/guide/device-containment.rst
garlick: This quirk probably belongs in an inline comment, if at all.

doc/guide/device-containment.rst
garlick: Oh, duh, maybe this should be under `options`.

doc/guide/device-containment.rst
garlick: Yeah that was the plan (easier to add/remove things there!)  I'll add a note.

garlick:
This is probably ready for another look.  It now reads `DevicePolicy` and `DeviceAllow` from the `options` dict in more or less the same form as the systemd properties of the same name.  In addition to the sharness tests, I've poked at it a bit on my test system and it seems to do what's advertised.

src/test/docker/checks/Dockerfile
```diff
@@ -48,11 +48,12 @@ RUN cd /tmp \
...
-     focal*) ;; \
-     el*) ;; \
-     fedora*) dnf install -y libubsan ;; \
-     *) (>&2 echo "Unknown BASE_IMAGE") ;; \
+     bionic*|focal*|jammy*|noble*|bookworm*) \
+         apt-get update && apt-get install -y linux-libc-dev ;; \
+     el*) yum install -y kernel-headers ;; \
+     fedora*) dnf install -y libubsan kernel-headers ;; \
+     alpine*) apk add --no-cache linux-headers ;; \
+     *) echo "Unknown BASE_IMAGE=$BASE_IMAGE" >&2 && exit 1 ;; \
```
grondo: We should just go ahead and get the testenv docker containers updated over in flux-core (just a simple PR), then when that's merged this commit can be dropped.

src/imp/exec/device.c
```diff
@@ -0,0 +1,128 @@
...
+        return -1;
+    }
+    if (kv_get (kv, "device.count", KV_INT64, &count) < 0) {
+        *dap = NULL;  /* absent == no containment */
+        return 0;
+    }
+    if (count < 0 || count > INT_MAX) {
+        errno = ERANGE;
+        return -1;
+    }
```
grondo: I might be more paranoid in the count check in decode here and use a more reasonable number than INT_MAX. Without this an attacker could set device.count to a large enough number to cause DoS by having the privileged process allocate a huge amount of memory.

src/imp/exec/device.c
```diff
@@ -0,0 +1,128 @@
...
+{
+    char type;
+    int major, minor;
+    char access[4];
+
+    if (sscanf (s, "%c:%d:%d:%3s", &type, &major, &minor, access) != 4
+        || (type != 'c' && type != 'b')) {
+        errno = EINVAL;
+        return -1;
+    }
```
grondo, line 354: `type` is validated here, should this call `access_is_valid (access)` as well?

grondo (dismissed):
Excellent! This LGTM! A few comments inline. We should probably get the testenv images updated so we don't need a future PR to revert the manual dependency install.

grondo (approved):
Re-approving since previous approval was cleared.

garlick:
Thanks!  setting mwp

PR #229, opened 2026-04-26
imp: add log-level TOML config key

garlick:
Problem: There is no way to enable IMP debug log output from the installed IMP executable.
    
Add a top-level `log-level` config key that accepts "warning", "info",  or "debug". When set, it overrides the default (info) level.


grondo (approved):
LGTM!

garlick:
Thanks, I"ll set MWP.

PR #234, opened 2026-07-24
imp: sanitize log messages

grondo:
The IMP's log functions format their arguments and pass the result to each output backend without filtering control characters. Some of those arguments could be attacker-influenced. Unfiltered, an embedded newline could forge a log line and an ESC byte could inject a terminal escape sequence into an administrator's view of the IMP's stderr/journal.

This is defense-in-depth: in practice the IMP is invoked only by the instance owner and most provided data is instance owner origin, so exposure is limited, but log output should never carry unsanitized untrusted bytes regardless of source.

The fix sanitizes in `vlog_msg()`, the single point every log message passes through before dispatch — replacing `bytes < 0x20` (except tab) and `0x7f` with `?` and stripping trailing newlines -- so every current and future call site is covered at once. Unit tests in `test/imp_log.c` verify the behavior.

This issue was found during a Claude Security scan of `src/imp`.

### DIFF
--- a/src/imp/imp_log.c
+++ b/src/imp/imp_log.c
@@ -171,6 +171,29 @@ int imp_log_set_level (const char *name, int level)
 /*
  *   Logging interface functions
  */
+
+/* Neutralize control characters in a fully-formatted log message before it
+ * is handed to any output backend. Without this, an embedded newline could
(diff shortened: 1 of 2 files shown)


garlick (approved):
LGTM!

grondo:
Thanks! I've set MWP.

